PR #2, opened 2023-03-11
setup cibuildwheel

tlambert03:
This PR would enable the running of cibuildwheel on a tagged commit, and upload wheels to PyPI

Note, that this would be unnecessary  if you merge #3


msschwartz21:
Closed in favor of #3

PR #5, opened 2023-03-11
Fix missing 't' for single nodes in track graph

adjavon:


src/cell_tracking_metrics/__init__.py
```diff
@@ -8,6 +8,3 @@
 
 __author__ = "Track Gals"
 __email__ = "draga.doncilapop1@monash.edu"
```
msschwartz21, line 10:
```suggestion
__email__ = "draga.doncilapop1@monash.edu"

from .tracking_data import TrackingData
from .tracking_graph import TrackingGraph

all = ["TrackingData", "TrackingGraph"]
```

src/cell_tracking_metrics/__init__.py
```diff
@@ -9,5 +9,5 @@
...
 
-from .tracking_data import TrackingData
-from .tracking_graph import TrackingGraph
+
+all = ["TrackingData", "TrackingGraph"]
```
msschwartz21, line 13:
```suggestion
__all__ = ["TrackingData", "TrackingGraph"]
```

PR #7, opened 2023-03-13
Add label_key to TrackingGraph

DragaDoncila:
Rather than requiring matchers to take the `label_key` as a parameter, store it on the `TrackingGraph` during loading.

### DIFF
--- a/src/cell_tracking_metrics/tracking_graph.py
+++ b/src/cell_tracking_metrics/tracking_graph.py
@@ -50,7 +50,8 @@ class TrackingGraph:
             frame_key (str, optional): The key on each node in graph that contains the time frame
                 of the node. Every node must have a value stored at this key. Defaults to 't'.
             label_key (str, optional): The key on each node that denotes the pixel value of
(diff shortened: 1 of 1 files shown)


src/cell_tracking_metrics/tracking_graph.py
```diff
@@ -35,20 +35,30 @@ class TrackingGraph:
...
         """A directed graph representing a tracking solution where edges go forward in time.
 
         Args:
             graph (networkx.DiGraph): A directed graph representing a tracking solution
                 where edges go forward in time.
             frame_key (str, optional): The key on each node in graph that contains the time frame
                 of the node. Every node must have a value stored at this key. Defaults to 't'.
+            label_key (str, optional): The key on each node that denotes the pixel value of
+                the node in the segmentation. Defaults to 'segmentation_id'. Pass `None` if
+                segmentation is not available.
```
msschwartz21, line 54:
```suggestion
                there is not a label attribute on the graph.
```

PR #34, opened 2023-04-11
Add report of matcher coverage

msschwartz21:
This addition to the base `Matched` class prints the fraction of matched GT and predicted nodes. Let me know if you think we should switch to a logger instead of printing, but I figured printing would makes sense since we print the results of the metrics later on.


DragaDoncila (approved):
Looks good imo, I think if we add a logger we can do that in one go. Thanks @msschwartz21

PR #35, opened 2023-06-05
Use appropriate label key for `gt` and `pred` graphs in `ctc` matching

DragaDoncila:
CTC matcher previously assumed `label_key` would be the same for `gt` and `pred` graphs. Obviously this may not be the case, so updating this.


msschwartz21 (approved):
LGTM!

PR #36, opened 2023-06-06
Remove detection matrix computation

DragaDoncila:
The original AOGM paper uses "detection matrices" to allow for quick error counting. However, because we are assigning `tp`/`fp` etc. attributes to each node, we don't benefit from this quick counting - we still have to iterate through the graph.

This PR removes the reliance on detection matrices and computes the errors directly on the mapping. This should also mean we can use the AOGM/CTC metric with other matchers, assuming they are one-to-one or many-to-one (can't do one-to-many).

### DIFF
--- a/examples/ctc.ipynb
+++ b/examples/ctc.ipynb
@@ -89,23 +89,23 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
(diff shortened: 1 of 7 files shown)


msschwartz21 (approved):
LGTM!

PR #37, opened 2023-08-04
Faster matchings/metrics

maweigert:
Thanks a lot for this nice package, its really great to finally have proper tracking metrics in python! 

I noticed that `run_metrics` is relatively slow for medium sized datasets (e.g. ~30s for the hela demo dataset) so this PR just add some small code changes that should improve the metrics calculations by a factor of 6-10x. 
Specifically:

- iou/overlap caluclation on slices rather than the full segmentations masks 
- vectorization of edge indexing 

With that the following 

```python 
from traccuracy import run_metrics
from traccuracy.loaders import load_ctc_data
from traccuracy.matchers import CTCMatched, IOUMatched
from traccuracy.metrics import CTCMetrics, DivisionMetrics
from timeit import default_timer


gt_data = load_ctc_data(
    'downloads/Fluo-N2DL-HeLa/01_GT/TRA',
    'downloads/Fluo-N2DL-HeLa/01_GT/TRA/man_track.txt'
)
pred_data = load_ctc_data(
    'sample-data/Fluo-N2DL-HeLa/01_RES', 
    'sample-data/Fluo-N2DL-HeLa/01_RES/res_track.txt'
)

t = default_timer()

ctc_results = run_metrics(
    gt_data=gt_data, 
    pred_data=pred_data, 
    matcher=CTCMatched, 
    metrics=[CTCMetrics, DivisionMetrics],
    metrics_kwargs=dict(
        frame_buffer=(0,1,2)
    )
)

t = default_timer() - t

print(f'took {t:.4} s')

```

gave me `3.2s` (new) vs `26.4s` (old)

### DIFF
--- a/src/traccuracy/matchers/_compute_overlap.py
+++ b/src/traccuracy/matchers/_compute_overlap.py
@@ -29,9 +29,8 @@ def get_labels_with_overlap(gt_frame, res_frame):
         overlapping_res_labels: List[int], labels of res boxes that overlap with gt boxes
         intersections: List[float], list of (intersection gt vs res) / (gt area)
     """
-    gt_frame, res_frame = gt_frame.astype(np.uint16, copy=False), res_frame.astype(
-        np.uint16, copy=False
-    )
+    gt_frame = gt_frame.astype(np.uint16, copy=False)
+    res_frame = res_frame.astype(np.uint16, copy=False)
     gt_props = regionprops(gt_frame)
     gt_boxes = [np.array(gt_prop.bbox) for gt_prop in gt_props]
     gt_boxes = np.array(gt_boxes).astype(np.float64)
--- a/src/traccuracy/matchers/_compute_overlap.py
+++ b/src/traccuracy/matchers/_compute_overlap.py
@@ -5,11 +5,13 @@ Written by Sergey Karayev
 Licensed under The MIT License [see LICENSE for details]
 Copyright (c) 2015 Microsoft
 """
+from typing import Tuple
+
 import numpy as np
 from skimage.measure import regionprops
 
 
-def _union_slice(a: tuple[slice], b: tuple[slice]):
+def _union_slice(a: Tuple[slice], b: Tuple[slice]):
     """returns the union of slice tuples a and b"""
     starts = tuple(min(_a.start, _b.start) for _a, _b in zip(a, b))
     stops = tuple(max(_a.stop, _b.stop) for _a, _b in zip(a, b))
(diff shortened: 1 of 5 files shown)


src/traccuracy/track_errors/_ctc.py
```diff
@@ -111,11 +112,16 @@ def get_edge_errors(gt_graph, comp_graph, node_mapping):
...
+    node_mapping_first = np.array([mp[0] for mp in node_mapping])
+    node_mapping_second = np.array([mp[1] for mp in node_mapping])
+
     # fp edges - edges in induced_graph that aren't in gt_graph
     for edge in tqdm(induced_graph.edges, "Evaluating FP edges"):
         source, target = edge[0], edge[1]
-        source_gt_id = list(filter(lambda mp: mp[1] == source, node_mapping))[0][0]
-        target_gt_id = list(filter(lambda mp: mp[1] == target, node_mapping))[0][0]
+
+        source_gt_id = node_mapping[np.where(node_mapping_second == source)[0][0]][0]
```
DragaDoncila, line 122: Just to double check my own understanding - we are still finding the corresponding gt/comp IDs one by one, but it's just that much faster using `np.where` on the prebuilt array?

src/traccuracy/matchers/_ctc.py
```diff
@@ -79,15 +79,16 @@ def _match_ctc(self):
...
+                intersection,
+            ) = get_labels_with_overlap(gt_frame, pred_frame)
+
             for i in range(len(overlapping_gt_labels)):
                 gt_label = overlapping_gt_labels[i]
                 pred_label = overlapping_pred_labels[i]
-                gt_blob_mask = gt_frame == gt_label
-                comp_blob_mask = pred_frame == pred_label
-                if detection_test(gt_blob_mask, comp_blob_mask):
+                if intersection[i] > 0.5:
```
DragaDoncila, line 91:
```suggestion
# CTC metrics only match comp IDs to a single GT ID if there is majority overlap
                if intersection[i] > 0.5:
```

Since we're not using the `detection_test` function anymore, it would be great to add a comment for the otherwise "magic" 0.5 value, and delete the `detection_test` function entirely.

src/traccuracy/matchers/_compute_overlap.py
```diff
@@ -9,6 +9,13 @@
 from skimage.measure import regionprops
 
 
+def _union_slice(a: tuple[slice], b: tuple[slice]):
```
DragaDoncila, line 12: Python 3.8 is complaining about typing here - could we replace with 

```python
from typing import Tuple

def _union_slice(a: Tuple[slice], b: Tuple[slice]):
```

I think that should stop it from complaining.

DragaDoncila (commented):
Wow @maweigert :exploding_head:  thank you for the killer PR! We knew there was more we could do to speed this up, but didn't realize just how easy it could be!

Just left two minor comments and a request for clarification for my own understanding. Super excited to get this merged :tada:

src/traccuracy/track_errors/_ctc.py
maweigert: Yes, seems like `filter` is way slower than the numpy intrinsics.

src/traccuracy/matchers/_compute_overlap.py
maweigert: Thx! Should be fixed

DragaDoncila:
Thanks for those updates @maweigert! Mergine now 🎉

Issue #43, opened 2023-09-07
Create and enforce set of valid graph annotation keys


cmalinmayor:
- [x] add Enum for valid node attrs and edge attrs in track_graph.py (make values be the strings we already have, so that the nx graph is inspectable)
- [x] `set_node_attribute` (and `set_edge_attribute`) only allows setting the valid node/edge attrs from the enum, and throws error otherwise (with descriptive message about adding attr to enum). If people need custom attrs, they add them to enum, which ensures no overlap of string representations.
- [x] update every place we annotate the graph to use the enums instead of hard coded strings
- [x] Check if frame key and location key overlap with enum strings upon initialization of TrackGraph

cmalinmayor:
Check if frame key and location key overlap with enum strings upon initialization of TrackGraph

msschwartz21:
Closed in #47 and #48

PR #46, opened 2023-09-07
Create Enums for our Node and Edge Attribute strings

cmalinmayor:
Addresses most of #43, except updating all references.


cmalinmayor:
Division tests are failing because we need to update references to `set_node_attributes` and `set_edge_attributes` to use the enums. This will happen on the div_org branch where other changes have already been made.

PR #47, opened 2023-09-07
Eliminate use of `TrackEvents` in in CTC error calculations

msschwartz21:
The `TrackEvents` class adds an additional layer of complication to metrics calculations that makes it harder to add additional calculations to the pipeline. This PR eliminates `TrackEvents` and instead just counts node and error attributes directly from the annotated graph. This also switches over the ctc related functions to use `TrackingGraph` methods instead of `networkx` when interacting with node and edge attributes.

Division related tests are failing but will be resolved by #48.


cmalinmayor (approved):
Looks good - minor changes were discussed and made during code review :)

PR #54, opened 2023-09-20
Annotate intertrack edges in the ctc `get_edge_errors` function

### LINKED ISSUE #45: `"is_intertrack_edge"` is used but not annotated on graphs in CTC errors
I noticed that we never annotate any edges as `"is_intertrack_edge"`, but the way this is written assumes that the attribute exists. 

https://github.com/Janelia-Trackathon-2023/traccuracy/blob/f8c982c57b8aa92ac8ab5410318ad2b20c0c6f76/src/traccuracy/track_errors/_ctc.py#L131-L132

In the test case, if I comment out the line where we set it on the input graph, the test fails.

https://github.com/Janelia-Trackathon-2023/traccuracy/blob/f8c982c57b8aa92ac8ab5410318ad2b20c0c6f76/tests/track_errors/test_ctc_errors.py#L76

Are we missing a bug here? @DragaDoncila can you take a look? I don't really understand the ctc errors well enough to make an informed evaluation.

msschwartz21:
This PR moves the annotation of intertrack edges into `get_edge_errors` based on the assumption that any edge connecting a parent to a daughter cells is an intertrack edge. This closes #45. 

A couple additional changes that emerged:
- `get_edge_errors` explicitly checks that node errors have been annotated and computes them if they are missing (since `get_edge_errors` requires access to true positive nodes)
- True positive edges are no longer annotated by `get_edge_errors`. The CTC metrics do not use the true positive annotation. Currently true positive edge and wrong semantic edge were considered mutually exclusive which Caroline and I both thought was a potentially dangerous assumption for someone who wasn't operating with CTC metrics. Without the true positive edges explicitly annotated, anyone who wants to count them can impose their own assumptions about what merits a true positive edge.


src/traccuracy/track_errors/_ctc.py
```diff
@@ -94,6 +98,17 @@ def get_edge_errors(matched_data: "Matched"):
     node_mapping_first = np.array([mp[0] for mp in node_mapping])
     node_mapping_second = np.array([mp[1] for mp in node_mapping])
 
+    # intertrack edges = connection between parent and daughter
+    for graph in [comp_graph, gt_graph]:
+        # Set to False by default
+        graph.set_edge_attribute(list(graph.edges()), EdgeAttr.INTERTRACK_EDGE, False)
+
+        for parent in graph.get_divisions():
```
DragaDoncila, line 106: @msschwartz21 we should probably also look at merges, which are also intertrack edges, so any edge whose target node has another incoming edge.

DragaDoncila (commented):
Hmm I'm not sure what's going on but on my dataset I'm getting 4 additional `ws_edges` on this branch compared to main. I don't have time to investigate right now what the difference is and which is right, but I'll have a look and get back to you.

src/traccuracy/track_errors/_ctc.py
```diff
@@ -80,12 +80,16 @@ def get_edge_errors(matched_data: "Matched"):
...
+    if not comp_graph.node_errors and not gt_graph.node_errors:
+        logger.warning("Node errors have not been annotated. Running node annotation.")
+        get_vertex_errors(matched_data)
+
     induced_graph = comp_graph.get_subgraph(
         comp_graph.get_nodes_with_flag(NodeAttr.TRUE_POS)
     ).graph
 
     comp_graph.set_edge_attribute(list(comp_graph.edges()), EdgeAttr.FALSE_POS, False)
-    comp_graph.set_edge_attribute(list(comp_graph.edges()), EdgeAttr.TRUE_POS, False)
```
DragaDoncila, line 88: @msschwartz21 I don't think I fully understand why we are not annotating true positive edges? You mention in the description that we can't rely on the assumption that `TP` and `WS` edges are mutually exclusive, but I think we can? I don't think a `TP` edge should ever be annotated with `WS`.

DragaDoncila:
> Hmm I'm not sure what's going on but on my dataset I'm getting 4 additional ws_edges on this branch compared to main.

I think it was a bug on my end, I don't see anything wrong with what we're doing here and the additional edges (in main, not here) shouldn't have been intertrack edges anyway. So just ignore this hehe

cmalinmayor:
@msschwartz21 Are we good to merge now?

msschwartz21:
I want to respond to Draga's question about true positive edges at some point, but it will take some additional thought on my part and shouldn't be a blocker to merging.

src/traccuracy/track_errors/_ctc.py
cmalinmayor: My stab at explaining this, that Morgan can elaborate on at her leisure: I admit that I am always confused by`WS` edges; however, it is my understanding that if there is a division and we correctly detect one of the edges to a daughter but not the other, the one we correctly detected would be labeled `WS` - correct? I would consider that a `TP` - we correctly identified the edge! Just because we missed the other daughter edge (a `FN` edge), that doesn't make the correctly recovered edge incorrect. In that way, something that is a `WS` edge in the CTC metric could be considered a `TP` elsewhere, making them not mutually exclusive.

cmalinmayor:
> I want to respond to Draga's question about true positive edges at some point, but it will take some additional thought on my part and shouldn't be a blocker to merging.

I took a stab at it, and you can follow up whenever you have a chance. I'll go ahead and merge then! Thanks :)

src/traccuracy/track_errors/_ctc.py
msschwartz21: ^ Caroline's description captures my thoughts perfectly

PR #55, opened 2023-09-20
README update

cmalinmayor:


cmalinmayor:
Still need to update API documentation, but README is much better now :)

Issue #57, opened 2023-09-22
Major slowdown for evaluating edges in current `main`


bentaculum:
* traccuracy version: `main` branch vs  `f8c982c57b8aa92ac8ab5410318ad2b20c0c6f76`
* Python version: 3.10.13
* Operating System: Ubuntu 22.04

### Description

The recent changes are considerably slowing down the evaluation of FP/FN edges. For the example below, one of the phase contrast datasets from the CTC, I get an estimate of 01:45h.  With commit `f8c982c57b8aa92ac8ab5410318ad2b20c0c6f76`, which I was using until just now, this evaluation gets done in only seconds.

### What I Did
I slightly adapted the [example](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/main/examples/ctc.ipynb), to simply compare GT vs GT of a slightly larger CTC dataset:

```python
import os
from pathlib import Path
import pprint
import urllib.request
import zipfile

from tqdm import tqdm

from traccuracy import run_metrics
from traccuracy.loaders import load_ctc_data
from traccuracy.matchers import CTCMatched, IOUMatched
from traccuracy.metrics import CTCMetrics, DivisionMetrics

pp = pprint.PrettyPrinter(indent=4)

url = "http://data.celltrackingchallenge.net/training-datasets/PhC-C2DL-PSC.zip"
data_dir = "downloads"

if not os.path.exists(data_dir):
    os.mkdir(data_dir)

filename = url.split("/")[-1]
file_path = os.path.join(data_dir, filename)
ds_name = filename.split(".")[0]

# Add a utility to make a progress bar when downloading the file
class DownloadProgressBar(tqdm):
    def update_to(self, b=1, bsize=1, tsize=None):
        if tsize is not None:
            self.total = tsize
        self.update(b * bsize - self.n)


if not os.path.exists(file_path):
    print(f"Downloading {ds_name} data from the CTC website")
    # Downloading data
    with DownloadProgressBar(
        unit="B", unit_scale=True, miniters=1, desc=url.split("/")[-1]
    ) as t:
        urllib.request.urlretrieve(url, file_path, reporthook=t.update_to)
    # Unzip the data
    # TODO add a progress bar to zip as well
    with zipfile.ZipFile(file_path, "r") as zip_ref:
        zip_ref.extractall(data_dir)


gt_data = load_ctc_data(
[...]

PR #59, opened 2023-09-22
Speed up CTC edge errors

bentaculum:
When calculating edge errors, there is a 1-to-1 mapping between computed graph nodes and GT graph nodes, see details below.

It is faster to use a dictionary for doing the node mapping with O(n log(n)) runtime, compared to repeatedly iterating over a list with np.where (O(n^2)).

Details: Potential 1-to-many matches, namely computed nodes that match multiple GT nodes (called non-split), are not part of the induced graph. Therefore, all edges incident to such nodes are also not part of the induced graph, leaving us with the desired 1-to-1 mapping for all nodes that are incident to existing edges in the induced graph, which we iterate over in the loop for finding FP edges.

Conversely, each GT node is only matched to at most one computed node, directly yielding the 1-1 matching for finding FN edges.

In case the test cases do not yet cover this edge case of a 1-to-many match for CTC we should probably add it.

### DIFF
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -181,15 +181,15 @@ class TrackingGraph:
                 Will raise KeyError if any of these node_ids are not present.
 
         Returns:
-            dict[hashable, dict]: A dictionary from node ids to node attributes
+            NodeView: Provides set-like operations on the nodes as well as node attribute lookup.
         """
         if limit_to is None:
             return self.graph.nodes
         else:
-            limited_nodes = {
-                _id: data for _id, data in self.graph.nodes if _id in limit_to
-            }
-            return limited_nodes
+            for node in limit_to:
+                if not self.graph.has_node(node):
+                    raise KeyError(f"Queried node {node} not present in graph.")
+            return self.graph.subgraph(limit_to).nodes
 
     def edges(self, limit_to=None):
         """Get all the edges in the graph, along with their attributes.
@@ -200,15 +200,16 @@ class TrackingGraph:
                 Will raise KeyError if any of these edge ids are not present.
 
         Returns:
-            dict[tuple[hashable], dict]: A dictionary from edge ids to edge attributes
+            OutEdgeView: Provides set-like operations on the edge-tuples as well as edge attribute
+                lookup.
         """
         if limit_to is None:
             return self.graph.edges
         else:
-            limited_edges = {
-                _id: data for _id, data in self.graph.edges if _id in limit_to
-            }
-            return limited_edges
+            for edge in limit_to:
+                if not self.graph.has_edge(*edge):
+                    raise KeyError(f"Queried edge {edge} not present in graph.")
+            return self.graph.edge_subgraph(limit_to).edges
 
     def get_nodes_in_frame(self, frame):
         """Get the node ids of all nodes in the given frame.
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -183,11 +183,12 @@ class TrackingGraph:
         Returns:
             dict[hashable, dict]: A dictionary from node ids to node attributes
         """
-        nodes = self.graph.nodes.items()
         if limit_to is None:
-            return dict(nodes)
+            return self.graph.nodes
         else:
-            limited_nodes = {_id: data for _id, data in nodes if _id in limit_to}
+            limited_nodes = {
+                _id: data for _id, data in self.graph.nodes if _id in limit_to
+            }
             return limited_nodes
 
     def edges(self, limit_to=None):
@@ -201,11 +202,12 @@ class TrackingGraph:
         Returns:
             dict[tuple[hashable], dict]: A dictionary from edge ids to edge attributes
         """
-        edges = self.graph.edges.items()
         if limit_to is None:
-            return dict(edges)
+            return self.graph.edges
         else:
-            limited_edges = {_id: data for _id, data in edges if _id in limit_to}
+            limited_edges = {
+                _id: data for _id, data in self.graph.edges if _id in limit_to
+            }
             return limited_edges
 
     def get_nodes_in_frame(self, frame):
--- a/src/traccuracy/track_errors/_ctc.py
+++ b/src/traccuracy/track_errors/_ctc.py
@@ -90,8 +90,12 @@ def get_edge_errors(matched_data: "Matched"):
     )
     gt_graph.set_edge_attribute(list(gt_graph.edges()), EdgeAttr.FALSE_NEG, False)
 
-    node_mapping_first = {mp[0]: mp[1] for mp in node_mapping}
-    node_mapping_second = {mp[1]: mp[0] for mp in node_mapping}
+    node_mapping_first = {
+        gt: comp for gt, comp in node_mapping if comp in induced_graph
+    }
+    node_mapping_second = {
+        comp: gt for gt, comp in node_mapping if comp in induced_graph
+    }
 
     # fp edges - edges in induced_graph that aren't in gt_graph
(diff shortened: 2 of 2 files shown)


bentaculum:
Little example: For the dataset in #57, getting FP/FN nodes now runs in less than a second for me, compared to ~10s in f8c982c57b8aa92ac8ab5410318ad2b20c0c6f76 and >1h in current `main`.

DragaDoncila (commented):
@bentaculum the logic for the edge errors being 1-to-1 is sound and I agree that iterating through the list is grim. Do you know which recent changes lead to the slowdown observed in `main` compared to after merging #37? I'm not getting the expected speed up in this branch - edge errors are being computed for me in similar times to on `main` - about 400x slower than with #37.

I noticed you create the dict mapping from the original matching - is this because we can only be certain of the 1-to-1ness after computing the induced graph? I think even if we do want 1-to-many and many-to-1 support, a dictionary mapping `node_key: set[node-key]` would still be much faster when vast majority of nodes are 1-to-1 anyway?

Would be nice to have some benchmarking in CI - we really want to be cautious of slowdowns.

bentaculum:
@DragaDoncila Thanks for the feedback.

In a fresh env I am not able to reproduce the speedup I reported either, sorry for the noise.

I noticed that the `TrackingGraph` currently converts the networkx edges stored as `EdgeView` to a dictionary each time we call `TrackingGraph.edges()` (and nodes similarly). This is very slow. Is there a reason why this is necessary? Maybe @msschwartz21 has some input on this as well.

Since the last sprint the track errors are stored as edge attributes on the `TrackingGraph`, which is the reason for the drastic slowdown.

I simply removed these two type casts, and tests still pass. Now I run `traccuracy.track_errors._ctc.get_edge_errors` in 3.3s for the example in #57.

For the `limit_to` attributes of `TrackingGraph.edges/nodes()` we might want to look into https://networkx.org/documentation/stable/reference/classes/generated/networkx.classes.graphviews.subgraph_view.html.

bentaculum:
> I noticed you create the dict mapping from the original matching - is this because we can only be certain of the 1-to-1ness after computing the induced graph? I think even if we do want 1-to-many and many-to-1 support, a dictionary mapping `node_key: set[node-key]` would still be much faster when vast majority of nodes are 1-to-1 anyway?

The original `Matched` object contains the 1-to-many matches. The dictionary that is currently created will store one arbitrary match for each 1-to-many matching, but since we iterate only over the edges of the induced graph this dictionary entry will never be used. So the `node_key: set[node-key]` is actually not necessary.

To make this logic more explicit, I now only add an entry to `node_mapping_first` and `node_mapping_second` if the node is part of the induced graph.

src/traccuracy/track_errors/_ctc.py
```diff
@@ -91,15 +90,19 @@ def get_edge_errors(matched_data: "Matched"):
     )
     gt_graph.set_edge_attribute(list(gt_graph.edges()), EdgeAttr.FALSE_NEG, False)
 
-    node_mapping_first = np.array([mp[0] for mp in node_mapping])
-    node_mapping_second = np.array([mp[1] for mp in node_mapping])
+    node_mapping_first = {
```
DragaDoncila, line 93: nit: would love to rename this to 'gt_comp_mapping' and `comp_gt_mapping` respectively

src/traccuracy/_tracking_graph.py
```diff
@@ -201,11 +202,12 @@ def edges(self, limit_to=None):
         Returns:
             dict[tuple[hashable], dict]: A dictionary from edge ids to edge attributes
         """
-        edges = self.graph.edges.items()
         if limit_to is None:
-            return dict(edges)
+            return self.graph.edges
         else:
-            limited_edges = {_id: data for _id, data in edges if _id in limit_to}
+            limited_edges = {
```
DragaDoncila, line 208: I think this could be `self.graph.edge_subgraph(limit_to).edges` and then the typing would be the same for both return statements.

DragaDoncila (approved):
Two small nits but otherwise love it - it's buttery smooth :tada: Thanks @bentaculum :heart:

src/traccuracy/_tracking_graph.py
bentaculum: Nice catch, put this in here 7e72be3. `networkx.DiGraph.edge_subgraph` and `networkx.DiGraph.subgraph` unfortunately don't throw an error when you ask for non-existing things. I do think that this is a reasonable check and added it. This way we also keep the API that you had before.

src/traccuracy/track_errors/_ctc.py
bentaculum: Very much agree, updated now.

src/traccuracy/_tracking_graph.py
bentaculum: I'm not sure about a good way to make docstrings that include classes from third-party packages, like networkx `OutEdgeView` here. Writing `networkx.classes.reportviews import OutEdgeView` in the docstring seems excessive.

Might be a good segway into typing traccuracy ;), or at least the core parts of it.

DragaDoncila: Yeah would really love to have this package better typed especially once we've sorta settled on the API

DragaDoncila (approved):
Love it! Thanks @bentaculum !🎉

cmalinmayor:
@bentaculum @DragaDoncila I do object to returning networkx specific objects in our track graph (e.g. NodeView, OutEdgeView). We want to be able to switch away from networkx and not change any of the API. Turning it into a dictionary is slow, I know, but perhaps we can do something else like return an iterator. Would that help with the speedup as well? 
FYI, we are also planning to add a way to access nodes/edges by attributes efficiently, essentially by storing a dictionary from attributes to node/edge ids, which should then negate the slowdown of storing attributes like FP only on the graph, isntead of in a separate counts object as well.

DragaDoncila:
@cmalinmayor thanks for flagging! I know initially we wanted to provide a full API via the `TrackingGraph`, so that the user would never have to dig into the networkx graph, but I'm not actually sure we do want to do that. For one thing, converting to and from the native networkx attributes is going to add some overhead even if it is efficient, and I'd like to avoid overhead wherever possible.

As I've been using `traccuracy`, I find myself only ever using the networkx API, rather than going through `TrackingGraph`. I think other users are also more likely to be familiar with the networkx API, and if they're not, then I would still say learning that is more useful (and easier, given the level of documentation), than learning our bespoke API. We could consider making our API exactly equivalent, but I think that's a lot off effort for not much gain.

I guess for me if I was to make any decision, it would be to further expose the underlying networkx graph, rather than add more functionality to the `TrackingGraph`.

cmalinmayor:
Re: the above comments and in the other thread here: https://github.com/Janelia-Trackathon-2023/traccuracy/pull/63#issuecomment-1760245852 @bentaculum @DragaDoncila 

Overall, I'm happy to go forward with this PR! However, I do think there are significant advantages to putting limitations on the networkx API in traccuracy. There are some things you can do in networkx that we explicitly want to ban (e.g. changing locations, adding/removing edges) - for power developers like you and Ben who fully understand networkx and traccuracy, you can easily avoid potential pitfalls, but as the user and/or developer base expands, it might be more important to enforce constraints. Additionally, if we retain control of the API we can make assumptions or augmentations to speed up access by attribute or location (linear on size of graph in native networkx, constant with the dictionaries I implemented in the other PR). Finally, if we ever have graphs too large to process in memory, we will want to implement or take advantage of an on-disk solution. I also just personally dislike the networkx API, but I know that's not a good reason to make everyone do a ton more work :).

The specific change in this PR is not introducing that big of a dependency, so I say we go forward with it but keep these potential issues with entangling the two libraries further in mind as we continue. If they never become issues, great, full steam ahead!

cmalinmayor (approved):

cmalinmayor:
Thanks @bentaculum!

PR #62, opened 2023-10-04
Add basic set of performance benchmarking tests

msschwartz21:
This PR introduces a preliminary workflow for monitoring performance as part of our CI. It uses `pytest-benchmark` and `github-action-benchmark`.

Currently it runs on the Fluo-ND2L-HeLa dataset with the sample solution graph that we have stored in `examples/sample-data` and measures the following:
- loading gt data
- loading predicted data
- IOU matching
- CTC matching
- CTC metrics
- Division metrics (on both iou matched and ctc matched)

@DragaDoncila @bentaculum @cmalinmayor let me know what you all think at least about `tests/bench.py`. I am considering this to be a pretty minimal first set of measurements just so that we have some sort of baseline to work off of, but let me know if you see anything else that we should add in this first iteration.

The testing workflow can be run locally with `pytest tests/bench.py`
Screen Shot 2023-10-04 at 3 55 04 PM

For the github action, I currently am only running the benchmarking on ubuntu-latest/python 3.11 for PRs. The action should eventually compare the results to the last commit on main and summarize any changes. It may take me a couple iterations of commits to main to get the reporting component working smoothly.

Additionally, a separate workflow will run on each commit pushed to main and publish plots to gh-pages ([example](https://benchmark-action.github.io/github-action-benchmark/dev/bench/)). This workflow may also take a couple commits to be fully functional.

### DIFF
--- a/.github/workflows/benchmark-report.yml
+++ b/.github/workflows/benchmark-report.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Run benchmark
         run: |
-          pytest tests/bench.py --benchmark-json output.json
+          pytest tests/bench.py --benchmark-min-rounds=1 --benchmark-json output.json
 
       - name: Store benchmark results
         uses: benchmark-action/github-action-benchmark@v1
--- a/.github/workflows/benchmark-report.yml
+++ b/.github/workflows/benchmark-report.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Run benchmark
         run: |
-          pytest tests/bench.py --benchmark-min-rounds=1 --benchmark-json output.json
+          pytest tests/bench.py --benchmark-json output.json
 
       - name: Store benchmark results
         uses: benchmark-action/github-action-benchmark@v1
--- a/.github/workflows/benchmark-report.yml
+++ b/.github/workflows/benchmark-report.yml
@@ -1,0 +1,43 @@
+# Do not run this workflow on pull request since this workflow has permission to modify contents.
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  # deployments permission to deploy GitHub pages website
+  deployments: write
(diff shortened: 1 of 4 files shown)


bentaculum:
Hi @msschwartz21,
thanks for setting this up, this is awesome! I have browsed over the changes in the github actions setup. I'm pretty new to this, but it looks good to me in general.
I understand that, to summarize:
- each time there is a merge into `main`, the benchmark gets automatically updated and pushed in json format.
- for a PR, the benchmark is run, and if there's at least a 2x slowdown for any of the benchmarks, a warning is thrown.
There might be other things that I have missed.

More thoughts:
- The color coding of the results table is confusing to me. It simply highlights the slower and faster tests based on absolute numbers, not very informative. I tried to turn it off, but couldn't find any suitable option.
- Right now, the benchmark takes more than a minute to run, notably because of the slow `test_ctc_metrics`. We might want to put in some hard thresholds using pytest-timeout with `@pytest.mark.timeout(10)`.

I made some small adjustments for speed, notably:
- pedantic call for more things
- decrease benchmark-min-rounds to 1.

msschwartz21:
Hey @bentaculum,

Thanks for looking this over and the good changes! I think your summary of the actions is correct although I'm still 🤞 that it works once merged into main. 

- I agree about the color coding, but similarly don't know how to change it. I think the comparison table that will be reported in PRs will be more informative.
- Good suggestion with `pytest-timeout`. I added it in just for the ctc metrics function for now, but we could consider adding a global timeout option in the future.
- I moved the configuration for `--benchmark-min-rounds` into `pyproject.toml` so that everyone will get the same behavior locally without need to remember to use the flag.

bentaculum (approved):
Went over this PR in quite some detail, looks good to me. Thanks again @msschwartz21, really cool to have this!

PR #63, opened 2023-10-05
Speed up accessing nodes/edges by attribute and node/edge attributes

### LINKED ISSUE #61: Speed up access of nodes/edges by attribute
Store a dictionary in the tracking graph from attributes to nodes/edges. Make sure it is updated when attributes are updated.

cmalinmayor:
Changes to `TrackingGraph` to improve performance related to attributes:
- Store dictionaries from NodeAttrs -> node ids and EdgeAttrs -> edge ids to efficiently access all nodes/edges with given attribute
- Use those dictionaries in `get_nodes_with_flag()`, `get_edges_with_flag()`, and `get_nodes_by_roi()`
- Speed up accessing node/edge attributes with specific functions `get_node_attribute()` and `get_edge_attribute()`, instead of having to go through `nodes()` and `edges()` dictionaries
- Speed up subgraph function by not re-computing all dictionaries from attrs to ids, but instead using intersection of sets of ids.
- Use `collections.Counter` in TrackingGraph test cases to make them order invariant
Closes #61

### DIFF
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -551,6 +551,56 @@ class TrackingGraph:
             else:
                 self.edges_by_flag[attr].discard(_id)
 
+    def get_node_attribute(self, _id, attr):
+        """Get the boolean value of a given attribute for a given node.
+
+        Args:
+            _id (hashable): node id
+            attr (NodeAttr): Node attribute to fetch the value of
+
+        Raises:
+            ValueError: if attr is not a NodeAttr
+
+        Returns:
+            bool: The value of the attribute for that node. If the attribute
+                is not present on the graph, the value is presumed False.
+        """
+        if not isinstance(attr, NodeAttr):
+            raise ValueError(
+                f"Provided attribute {attr} is not of type NodeAttr. "
+                "Please use the enum instead of passing string values, "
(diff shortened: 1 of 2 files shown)


msschwartz21 (approved):
LGTM!

cmalinmayor:
@msschwartz21 Did the benchmarking run on this PR? How do I tell?

msschwartz21:
No the benchmarking isn't running on this yet. I will need to get #62 merged in first, but eventually it should comment on the PR with a summary of the benchmarking.

bentaculum:
Now that the benchmark ran, `test_ctc_metrics` is still taking 157s, vs 1s in the networkx-based draft I wrote #59. While I agree that it would be nice to not have dependencies on an external API, to me making use of optimized data structures and objects like for example in networkx/rustworkx, and defining an API for that, is preferable to writing our own data structures to storing the graph.

bentaculum:
In the case of networkx, many of it's useful view objects support python-dictionary-style access, which we could easily make use of, probably in a wrapper class/inherited class of `TrackingGraph` to make it clean.

DragaDoncila:
@bentaculum I tend to agree with you here. The more I use traccuracy the more I want our data structures to be a very very thin layer on top of networkx

PR #64, opened 2023-10-12
Improve the benchmarking workflow to generate a commit comment on PRs

msschwartz21:
I couldn't get the `github-actions-benchmark` to perform the way I wanted for PRs (although it's working great for generating the charts on github pages). Instead I wrote a slightly more manual action that I think gives us what we want.

The benchmarking action triggered for each PR now does the following:
- Checks if benchmarks have been cached for the base commit and computes and saves them if not
- Runs benchmarking for the current commit 
- Generates a table comparing the mean runtime for head and base with the percent change
- Comments on the commit with the comparison table

@bentaculum are there any other statistics that you would want to see in that table?

### DIFF
--- a/.github/workflows/benchmark-pr.py
+++ b/.github/workflows/benchmark-pr.py
@@ -22,9 +22,14 @@ def make_report(old_path, new_path, out_file):
     new = load_stats(new_path)
 
     # Merge on benchmark name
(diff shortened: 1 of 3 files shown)


bentaculum:
Looks great, just added some small changes to the autogenerated comment table, and added the tabulate dep.

bentaculum:
Oh, thanks for figuring this out, I was quite baffled why this didn't go through.

PR #70, opened 2023-10-30
style: update pre-commit

tlambert03:
This pr runs `pre-commit autoupdate`, to bring the linting deps up to date... and (all of these are optionally and can be removed if you like):

1. adds [the `TID` tidy-imports rule](https://docs.astral.sh/ruff/rules/#flake8-tidy-imports-tid), which enforces the absolute import patterns you've generally been following here.
2. adds [the `TCH` type-checking rule](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tch) which nicely catches cases where an import should be removed from, or moved to, an `if TYPE_CHECKING` block (it can autofix it too)
3. adds https://github.com/crate-ci/typos to the checks, which is a nice, fast, rust-based code-spelling checker (it caught and fixed few issues in the docstrings, also included here)

closes #33

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,24 +4,29 @@ ci:
   autoupdate_commit_msg: "ci(pre-commit.ci): autoupdate"
 
 repos:
-  - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.254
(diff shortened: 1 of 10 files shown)


msschwartz21 (approved):
LGTM! Big fan of the spell checker 😂

PR #103, opened 2023-11-03
Refactor existing `Matched` into separate `Matcher` which returns `Matched`

### LINKED ISSUE #86: API Update: Matcher class or function that produces Matched object
Right now, the Matched object serves as the data container and as the thing that does the computing. We want to change this so that there is a Matcher class (or function?) that does the computing and a Matched object that contains the data that comes out of the matching.

### LINKED ISSUE #101: [BUG] IOUMatched with various IoU thresholds not reflected in metrics
Experimenting with different values of IoU threshold, the performance is always the same. This is true even when the IoU is set to 1.0 which should discard virtually all cells 👇 

_Note:_ As per @cmalinmayor's suggestion, I changed both `division_annotations` flags in GT and pred data to False to encourage re-computation. This didn't help 😞 

```
IoU Threshold : 1.0
Matching frames: 100%|██████████| 92/92 [00:05<00:00, 16.68it/s]
1.0
{   'Frame Buffer 0': {   'Division F1': 0.4318181818181818,
                          'Division Precision': 0.4175824175824176,
                          'Division Recall': 0.4470588235294118,
                          'False Negative Divisions': 94,
                          'False Positive Divisions': 106,
                          'Mitotic Branching Correctness': 0.2753623188405797,
                          'True Positive Divisions': 76},
    'Frame Buffer 1': {   'Division F1': 0.4318181818181818,
                          'Division Precision': 0.4175824175824176,
                          'Division Recall': 0.4470588235294118,
                          'False Negative Divisions': 94,
                          'False Positive Divisions': 106,
                          'Mitotic Branching Correctness': 0.2753623188405797,
                          'True Positive Divisions': 76}}
```
vs.
```
IoU Threshold : 0.7
Matching frames: 100%|██████████| 92/92 [00:05<00:00, 15.83it/s]
0.7
{   'Frame Buffer 0': {   'Division F1': 0.4318181818181818,
                          'Division Precision': 0.4175824175824176,
                          'Division Recall': 0.4470588235294118,
                          'False Negative Divisions': 94,
                          'False Positive Divisions': 106,
                          'Mitotic Branching Correctness': 0.2753623188405797,
                          'True Positive Divisions': 76},
    'Frame Buffer 1': {   'Division F1': 0.4318181818181818,
                          'Division Precision': 0.4175824175824176,
                          'Division Recall': 0.4470588235294118,
                          'False Negative Divisions': 94,
                          'False Positive Divisions': 106,
                          'Mitotic Branching Correctness': 0.2753623188405797,
                          'True Positive Divisions': 76}}
```

msschwartz21:
This PR closes #86 by changing our current `Matched` class into a `Matcher` which is initialized only with kwargs. A subsequent call of `Matcher().compute_mapping(gt_graph, pred_graph)` returns a `Matched` data object. 

Other things to note:
- Each `Matched` object creates a copy of the associated gt and pred tracking graphs which closes #101 
- I decided not to change the type of `mapping` to dictionaries because it causes issues when we have a mapping that is not one-to-one

@DragaDoncila or @bentaculum can one of you review this PR since Caroline and Kristina were part of the hacking session?

### DIFF
--- a/src/traccuracy/track_errors/_ctc.py
+++ b/src/traccuracy/track_errors/_ctc.py
@@ -26,7 +26,7 @@ def get_vertex_errors(matched_data: "Matched"):
 
     Parameters
     ----------
-    matched_data: Matched
+    matched_data: traccuracy.matchers.Matched
         Matched data object containing gt and pred graphs with their associated mapping
     """
     comp_graph = matched_data.pred_graph


src/traccuracy/matchers/_matched.py
```diff
@@ -1,45 +1,82 @@
...
+        gt_graph (TrackingGraph): Tracking graph object for the gt
+        pred_graph (TrackingGraph): Tracking graph object for the pred
+
+    """
+
+    def __init__(
+        self,
+        gt_graph: "TrackingGraph",
+        pred_graph: "TrackingGraph",
+        mapping: list[tuple[Any, Any]],
```
msschwartz21, line 78: The typing part of this line is causing the tests to fail in Python 3.8

@DragaDoncila @tlambert03 help please 😬 typing is not my strong suit

tlambert03: before 3.9, `tuple` (the actual symbol itself representing the python type) was not subscriptable.  meaning there was no `__class_getitem__` on the `tuple` class, and you couldn't do `tuple[...]`.  There are two options to deal with that:

1. use `typing.Tuple` instead of `tuple`:
   ```python
   from typing import Tuple, List
   # ...
   mapping: List[Tuple[Any, Any]]
   ```
2. add `from __future__ import annotations` to the top of the file... then use the regular types.
    ```python
     from __future__ import annotations
     # ...
     mapping: list[tuple[Any, Any]]
     ```

    The implication of adding `from __future__ import annotations` is that python treats all of your type hints as strings, and does not attempt to evaluate them at runtime (so, it doesn't matter that tuple doesn't have a `__class_getitem__`, method since it won't be called at runtime anyway).  details in [PEP 563](https://peps.python.org/pep-0563/)

In general, I prefer the second option (`__future__.annotations`) since it makes for cleaner looking modern files.  However, there are a few cases where that can be problematic; namely, if you or any other libraries try to evaluate your type hints at runtime (such as pydantic, typer, fastAPI, or magicgui might do)... so put a pin in the general idea that you might run into this later

src/traccuracy/matchers/_matched.py
```diff
@@ -1,45 +1,82 @@
...
+
+class Matched:
+    """Matched data class which stores TrackingGraph objects for gt and pred
+    and the computed mapping
+
+    Each TrackingGraph will be a new copy on the original object
+
+    Args:
+        gt_graph (TrackingGraph): Tracking graph object for the gt
+        pred_graph (TrackingGraph): Tracking graph object for the pred
```
tlambert03, line 70: note also, you're getting some errors in the docs build as well:

```
/home/docs/checkouts/readthedocs.org/user_builds/traccuracy/checkouts/103/docs/source/autoapi/traccuracy/matchers/_matched/index.rst:46: WARNING: more than one target found for cross-reference 'TrackingGraph': traccuracy._tracking_graph.TrackingGraph, traccuracy.TrackingGraph
/home/docs/checkouts/readthedocs.org/user_builds/traccuracy/checkouts/103/docs/source/autoapi/traccuracy/matchers/_matched/index.rst:46: WARNING: more than one target found for cross-reference 'TrackingGraph': traccuracy._tracking_graph.TrackingGraph, traccuracy.TrackingGraph
/home/docs/checkouts/readthedocs.org/user_builds/traccuracy/checkouts/103/docs/source/autoapi/traccuracy/matchers/_matched/index.rst:63: WARNING: more than one target found for cross-reference 'TrackingGraph': traccuracy._tracking_graph.TrackingGraph, traccuracy.TrackingGraph
/home/docs/checkouts/readthedocs.org/user_builds/traccuracy/checkouts/103/docs/source/autoapi/traccuracy/matchers/_matched/index.rst:63: WARNING: more than one target found for cross-reference 'TrackingGraph': traccuracy._tracking_graph.TrackingGraph, traccuracy.TrackingGraph
```

and I think it can be fixed by changing this to:

```suggestion
        gt_graph (traccuracy.TrackingGraph): Tracking graph object for the gt
        pred_graph (traccuracy.TrackingGraph .TrackingGraph): Tracking graph object for the pred
```

msschwartz21: Switching to "traccuracy.TrackingGraph" fixed the problem with the docs build, but I noticed in several other files we follow the same pattern of just using "TrackingGraph" in the docstring without any issues. Any idea why?

tlambert03: > in several other files we follow the same pattern of just using "TrackingGraph" in the docstring without any issues

in the docstring?  or in the type hint?  this is somewhat tricky stuff.  For docstrings it depends on how sphinx is doing the type resolution from string to symbol.  just for the sake of debugging... and to clarify whether we're talking about docstrings or type hints, can you point to the module you noticed "TrackingGraph"

msschwartz21: Realistically as long as it's working I can live with the ambiguity 🤷 But here's the places I noticed it:
https://github.com/Janelia-Trackathon-2023/traccuracy/blob/7b4156280613216c70c7729026c5617912d609e3/src/traccuracy/matchers/_iou.py#L116-L117
https://github.com/Janelia-Trackathon-2023/traccuracy/blob/7b4156280613216c70c7729026c5617912d609e3/src/traccuracy/matchers/_ctc.py#L31-L32

tlambert03: oh hell, those are indeed super annoying :joy:  I have no explanation!  stuck in the mysteries of sphinx at the moment.  I too am happy to live with the ambiguity if you are :)

cmalinmayor:
@DragaDoncila @bentaculum Do either of you have the chance to take a quick look at this and #109? Morgan and I are blocked by these not being merged, and I think they are fairly straightforward implementations of things we discussed if you have 10 minutes.

bentaculum (commented):
Great, looks good to me apart from the small comments I left. I didn't check out the tests yet, I'm assuming they were just updated according to the new API, right? Or also a test for the bug that this started off with?

src/traccuracy/matchers/_matched.py
```diff
@@ -1,45 +1,88 @@
...
-    from traccuracy._tracking_graph import TrackingGraph
+from traccuracy._tracking_graph import TrackingGraph
 
 logger = logging.getLogger(__name__)
 
 
-class Matched(ABC):
-    def __init__(self, gt_graph: "TrackingGraph", pred_graph: "TrackingGraph"):
-        """Matched class which takes TrackingData objects for gt and pred, and computes matching.
+class Matcher(ABC):
```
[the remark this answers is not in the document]
msschwartz21: In metrics we have a `_base.py` file instead which would save us from an additional file in this case. Do you have a preference?

src/traccuracy/matchers/_iou.py
```diff
@@ -98,29 +98,36 @@ def match_iou(gt, pred, threshold=0.6):
...
+            Matched: Matched data object containing IOU mapping
+        """
         # Check that segmentations exist in the data
         if gt_graph.segmentation is None or pred_graph.segmentation is None:
             raise ValueError(
                 "Segmentation data must be provided for both gt and pred data"
             )
 
-        super().__init__(gt_graph, pred_graph)
+        mapping = match_iou(gt_graph, pred_graph, threshold=self.iou_threshold)
```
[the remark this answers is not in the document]
msschwartz21: Would you get rid of the copy in the `Matched` class? My only hesitation is keeping it in the Matched class ensures that it will always happen even if we have someone implementing a new `Matcher`. I'm worried about making two sets of copies as we get into larger graphs.

msschwartz21:
> I didn't check out the tests yet, I'm assuming they were just updated according to the new API, right? Or also a test for the bug that this started off with?

The changes to the tests are only for the new API.

src/traccuracy/matchers/_matched.py
```diff
@@ -1,45 +1,88 @@
...
+class Matched:
+    """Matched data class which stores TrackingGraph objects for gt and pred
+    and the computed mapping
+
+    Each TrackingGraph will be a new copy on the original object
+
+    Args:
+        gt_graph (traccuracy.TrackingGraph): Tracking graph object for the gt
+        pred_graph (traccuracy.TrackingGraph): Tracking graph object for the pred
+
```
cmalinmayor: No mapping in docstring 😞

src/traccuracy/matchers/_base.py
```diff
@@ -0,0 +1,88 @@
...
+            ValueError: gt and pred must be a TrackingGraph object
+        """
+        if not isinstance(gt_graph, TrackingGraph) or not isinstance(
+            pred_graph, TrackingGraph
+        ):
+            raise ValueError(
+                "Input data must be a TrackingData object with a graph and segmentations"
+            )
+
+        matched = self._compute_mapping(gt_graph, pred_graph)
```
msschwartz21, line 44: Copying should be moved up here

src/traccuracy/matchers/_iou.py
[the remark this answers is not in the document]
msschwartz21: I found a place to move it into the `Matcher` class which I think makes the most sense.

src/traccuracy/matchers/_matched.py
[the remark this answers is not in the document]
bentaculum: Sounds good.

src/traccuracy/matchers/_iou.py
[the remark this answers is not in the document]
bentaculum: Thanks, lgtm.

bentaculum (approved):

src/traccuracy/matchers/_base.py
```diff
@@ -0,0 +1,92 @@
...
+            NotImplementedError
+        """
+        raise NotImplementedError
+
+
+class Matched:
+    """Matched data class which stores TrackingGraph objects for gt and pred
+    and the computed mapping
+
+    Each TrackingGraph will be a new copy on the original object
```
cmalinmayor, line 75: Typo: `a new copy on the original object` -> `a new copy of the original object`

msschwartz21:
@tlambert03 Can you weigh in on best practices for imports? @cmalinmayor and I discussed and changed to importing `Matched` in `traccuracy.matchers.__init__`. This seems to have the benefit of making a lot of imports cleaner (`from traccuracy.matchers import Matched` vs. `from traccuracy.matchers._base import Matched`). However, it's raising questions about when should we use relative imports as opposed to absolute imports and some concerns about creating circular imports.

tlambert03:
yeah i think it looks better and I'd go with it too.  4cf32e056b9225245d6c31fb84d15287120512b2 itself looks just fine, no fears about circular imports there.  (and, ultimately, if you don't get any in your tests, you're fine.)

> it's raising questions about when should we use relative imports as opposed to absolute imports and some concerns about creating circular imports.

I'm not 100% sure about this, but I *think* that the exact syntax of your import won't matter here.  that is, i think that python will do the same thing in these two lines (and neither would be safer in terms of relative imports)

```python
# inside of a/b/c/d.py
from a.b.c.e import E
from .e import E
```

i think the change looks good 👍

cmalinmayor:
Thanks @tlambert03! Can you also enlighten us on why the docs build is breaking? I suppose the error is pretty clear, it found the Matched class in two locations, but how do we fix it?

msschwartz21:
> Thanks @tlambert03! Can you also enlighten us on why the docs build is breaking? I suppose the error is pretty clear, it found the Matched class in two locations, but how do we fix it?

I've got this one. It came up in another PR so I know how to fix it.

tlambert03:
If that strict docs build ever gets annoying, feel free to disable it.  I find that if you ignore docs warnings, they gradually build up and get unmanageable over time... but it can also be annoying to force yourself to deal with them as well :)

cmalinmayor (approved):
I think we can (finally) merge! :)

PR #104, opened 2023-11-03
style: cleanup string type annotations

tlambert03:
little PR that cleans up type annotations in some places.  main points here are:

1. as you've already done here, wrapping a type hint in quotes `"TrackingGraph"` avoids needing to evaluate that symbol at runtime, so you can guard the import behind an `if TYPE_CHECKING` clause.  As a convenience, and if you find that you are doing that for many/most of the hints in a given module, you can simply add `from __future__ import annotations` to the top; it's the same as wrapping everything in a string.
1. whenever you use a type from `builtins` (like `int`, or `float`, etc...), you never need to wrap it in strings.
1. note that while i did remove a number of strings from `cli.py` (for builtin types), I did *not* add `from __future__ import annotations` to the top of the file, because those type hints need to be evaluated at runtime by typer.

also... probably don't merge this before #103, cause of conflicts.  I'll clean up after you merge that


msschwartz21 (approved):
Thanks Talley! So much nicer to read now :)

Issue #87, opened 2023-10-31
API Update: Instantiate Matcher and Metrics classes before passing into run_metrics


cmalinmayor:
Instantiating Matcher and Metrics classes before passing into run_metrics function avoids the kwargs arguments, which are fairly messy. This requires changing the API of the matcher to pass the data into the `compute` method instead of into the constructor. Make this change concurrently with or after #86 .

msschwartz21:
Instead of returning a list of tuples return two dictionaries that provides lookup between nodes in either direction.

PR #109, opened 2023-11-04
Refactor `Metrics` class to pass data into compute method not constructor

msschwartz21:
In support of addressing #87. I'll make the final changes to `run_metrics` in another PR once this one and #103 are merged.


src/traccuracy/metrics/_divisions.py
```diff
@@ -32,45 +32,50 @@
...
 
+from typing import TYPE_CHECKING
 
 from traccuracy._tracking_graph import NodeAttr
 from traccuracy.track_errors.divisions import _evaluate_division_events
 
 from ._base import Metric
 
+if TYPE_CHECKING:
+    from ._base import Matched
```
cmalinmayor, line 45: This import isn't right, is it? It implies that `Matched` is in the metrics/_base.py

src/traccuracy/metrics/_base.py
```diff
@@ -6,32 +8,22 @@
...
-        Each current matching method will be a subclass of Matched e.g. CTCMatched or IOUMatched.
-        The Matched objects will store both gt and pred data, as well as the mapping,
-        and any additional private attributes that may be needed/used e.g. detection matrices.
-        Metric subclasses will take keyword arguments to set the weights of various error counts.
-
-        Args:
-            matched_data (Matched): Matched object for set of GT and Pred data
-        """
-        self.data = matched_data
-        self.results = self.compute()
```
cmalinmayor, line 27: I like this change as well - it makes sense that the Metric wouldn't store the results if it doesn't store the data.

cmalinmayor (commented):
Overall looks great! Had one question about an import (although I assume the tests would have broken if my understanding was correct, so it is probably just me being confused). We need to standardize the type annotations with/without quotes and/or `from __future__ import annotations` but we can do that in #104

src/traccuracy/metrics/_divisions.py
msschwartz21: Yeah you're right. I thought I had fixed that in the matcher PR, but maybe the change got lost in the merge.

msschwartz21: I think since we import `Matched` in `metrics/_base.py` it is exposed for subsequent import here. But it's super sketch

cmalinmayor (approved):
Excellent! :) Thanks Morgan!

Issue #110, opened 2023-11-04
API Update: TrackingGraph.set_node_attribute and set_edge_attribute are overloaded


cmalinmayor:
I tried to make `TrackingGraph.set_node_attribute` and `TrackingGraph.set_edge_attribute` generic so that users could pass in either an id or a list of ids, with the hope of automatically detecting which it is. The way it is implemented now, the function literally checks for an instance of a list, so that if people have iterables of other types they need to cast them to list for it to be detected. This is clunky and weird (and potentially slow). However, because node_ids can be any hashable type, including string, the usual pythonic "try iterating and see if it works" is not effective for testing for lists of ids.

I think we need to be explicit and split each function into two functions, one for updating a single node/edge and another for updating a set of nodes/edges. So far, we only update a single node/edge and all nodes/edges, so perhaps `set_node_attribute` (for a single node) and `set_all_node_attributes` covers all our use cases (with the analagous edge attribute functions as well).

Issue #89, opened 2023-10-31
API Update: Identify core TrackingGraph functions, eliminate unnecessary code


cmalinmayor:
Go through the TrackingGraph API and:
- [x] remove functions that are not used in any metrics
- [x] remove keyword args that are not used in any metrics
- [x] add typing to the functions

Go through the metrics and:
- [ ] identify access patterns in the metrics that are not well served by the current API
- [ ] add to or adapt current API to address those access patterns

Finally:
- [ ] #135

Issue #91, opened 2023-10-31
Make explicit (optional) validation function in TrackingGraph constructor


cmalinmayor:
This will be more efficient for power users who know the networkx graph is correct, and for creating subgraphs of the TrackingGraph. Should probably put the validation in a separate function and call it or not based on the argument.

cmalinmayor:
I just ran into the issue that the label_key was incorrect. We should check  that if there is a segmentation provided, every node has the label_key in the validation as well to avoid this issue (or throw a warning if we don't want to be strict). The actual error message I got was a KeyError in the CTCMatcher which was very uninformative.

PR #111, opened 2023-11-04
Prune tracking graph API 

cmalinmayor:
This is a first attempt at streamlining the TrackingGraph API as mentioned in #89. See commit messages for details of my thought process.

Questions I want feedback on include:
 1. Now that I have removed the `limit_to` args in `TrackingGraph.nodes` and `TrackingGraph.edges`, these functions literally just wrap `networkx.nodes` and `networkx.edges`. Additionally, their usage is very limited in the whole library. E.g. [getting the number of nodes](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/27da5a762328bd383d3242469794351b3aa1fb17/src/traccuracy/matchers/_matched.py#L29) for which we could implement `__len__` and [setting a NodeAttr to True for all nodes](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/27da5a762328bd383d3242469794351b3aa1fb17/src/traccuracy/track_errors/_ctc.py#L44) which needs to be changed anyways as described in #110. Shall we remove them?
 2. The use of `get_nodes_with_attribute` in the IOU matcher [here](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/27da5a762328bd383d3242469794351b3aa1fb17/src/traccuracy/matchers/_iou.py#L87) is essentially to find the node with the given label_id in the given time point so we can construct the matching tuples of node_ids. Thoughts on constructing a dictionary {time: {segmentation_id : node_id}} (technically two dictionaries, one for gt and one for pred) before we enter the for loop to improve efficiency and remove this function?
 3. `TrackingGraph.get_node_attribute` and `TrackingGraph.get_edge_attribute` making the assumption that not having a flag means the attribute is negative. Since most of our attributes are rare (exception being True Positive, which we technically don't need to annotate), this saves time setting lots of attributes to False. However, it does lose out on explicitness of the annotations. Would we prefer explicitly annotating every attribute on every node/edge?
 4. Shortening `attribute` to `attr` everywhere? Also, flag and attribute are now basically synonymous, since all our supported attributes are True/False. Should we pick one? Use Flag for our reserved ones and Attribute for custom annotations that can be added directly through networkx? General nomenclature thoughts welcome.


src/traccuracy/_tracking_graph.py
```diff
@@ -201,44 +201,22 @@ def __init__(
         self.node_errors = False
         self.edge_errors = False
 
-    def nodes(self, limit_to=None):
+    def nodes(self):
```
msschwartz21, line 204: Can we change `nodes` and `edges` to `@property` so that they aren't callable and match the networkx behavior? Of course at that point it's close to just justifying getting rid of them entirely but constantly having to refer to the graph attribute gets ugly.

msschwartz21:
>1. Now that I have removed the limit_to args in TrackingGraph.nodes and TrackingGraph.edges, these functions literally just wrap networkx.nodes and networkx.edges. Additionally, their usage is very limited in the whole library. E.g. [getting the number of nodes](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/27da5a762328bd383d3242469794351b3aa1fb17/src/traccuracy/matchers/_matched.py#L29) for which we could implement __len__ and [setting a NodeAttr to True for all nodes](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/27da5a762328bd383d3242469794351b3aa1fb17/src/traccuracy/track_errors/_ctc.py#L44) which needs to be changed anyways as described in https://github.com/Janelia-Trackathon-2023/traccuracy/issues/110. Shall we remove them?

I vote for changing nodes and edges to properties so that they mirror the networkx api but allow us to avoid the extra `.graph` calls everywhere.

>2. The use of get_nodes_with_attribute in the IOU matcher [here](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/27da5a762328bd383d3242469794351b3aa1fb17/src/traccuracy/matchers/_iou.py#L87) is essentially to find the node with the given label_id in the given time point so we can construct the matching tuples of node_ids. Thoughts on constructing a dictionary {time: {segmentation_id : node_id}} (technically two dictionaries, one for gt and one for pred) before we enter the for loop to improve efficiency and remove this function?

I'm in favor of removing the function. Agnostic to how the change is implemented in the iou matcher.

>3. TrackingGraph.get_node_attribute and TrackingGraph.get_edge_attribute making the assumption that not having a flag means the attribute is negative. Since most of our attributes are rare (exception being True Positive, which we technically don't need to annotate), this saves time setting lots of attributes to False. However, it does lose out on explicitness of the annotations. Would we prefer explicitly annotating every attribute on every node/edge?

I don't love the pattern of setting attributes to False by default and then flipping them to True later. It seems like it adds a lot of clutter to the graph. Assuming we stop annotating False attributes, having a function that explicitly encodes the assumption that absent annotations are False seems like a good idea to prevent mistakes down the road.

>4. Shortening attribute to attr everywhere? Also, flag and attribute are now basically synonymous, since all our supported attributes are True/False. Should we pick one? Use Flag for our reserved ones and Attribute for custom annotations that can be added directly through networkx? General nomenclature thoughts welcome.

I like flag for our special attributes and attribute for other annotations. Flag is a bit nicer to look at than attr

msschwartz21:
Consider changing `get_nodes_with_flags` to `get_nodes_by_flag`

src/traccuracy/_tracking_graph.py
DragaDoncila: Yep, agreed @msschwartz21

DragaDoncila (commented):
> I don't love the pattern of setting attributes to False by default and then flipping them to True later. It seems like it adds a lot of clutter to the graph. Assuming we stop annotating False attributes, having a function that explicitly encodes the assumption that absent annotations are False seems like a good idea to prevent mistakes down the road.

I actually think it's dangerous to maintain ragged attributes on our graphs. Not only is it likely to lead to confusing `KeyErrors` down the track for users who grab the networkx graph and use it for downstream processing, but it might also lead someone to believe that a graph has had its errors computed when it hasn't (they query `is_fp` for a node and get a response back so they think errors are annotated). It's also inexplicit which usually leads to confusion down the line. Setting an attribute to a default value and then overriding it as required is a pretty common pattern, and I think it's what we need here. 

I also don't think the `get_node_attribute` and `get_edge_attribute` functions should return `False` when the attribute doesn't exist, they should raise an error instead. Also nit, but if these two functions only work for the boolean error attributes, I think that should be reflected in the name of the function. Something like `get_node_error_flag` or something?  

> Thoughts on constructing a dictionary {time: {segmentation_id : node_id}} (technically two dictionaries, one for gt and one for pred) before we enter the for loop to improve efficiency and remove this function?

Big +1. We should definitely avoid trawling through the list of nodes wherever possible. Especially for matching, we should ideally be finding the correct node in `O(1)` since we're going to eventually look for all nodes, giving us quadratic growth otherwise.

cmalinmayor:
> I actually think it's dangerous to maintain ragged attributes on our graphs. Not only is it likely to lead to confusing `KeyErrors` down the track for users who grab the networkx graph and use it for downstream processing, but it might also lead someone to believe that a graph has had its errors computed when it hasn't (they query `is_fp` for a node and get a `False` back so they think errors are annotated). It's also inexplicit which usually leads to confusion down the line. Setting an attribute to a default value and then overriding it as required is a pretty common pattern, and I think it's what we need here.

Thanks for the feedback! I'll revert the change and add some documentation regarding this convention.

cmalinmayor:
@DragaDoncila @msschwartz21 I think this is ready for review now. Most of the non-TrackingGraph code changes come from the following relatively minor changes:
- Renaming `NodeAttr`/`EdgeAttr` to `NodeFlag`/`EdgeFlag`
- Refactoring `set_node/edge_attribute` to `set_flag_on_node/edge` and `set_flag_on_all_nodes/edges`

There is also the IOU matcher refactoring which is a more substantive change but necessary to remove `TrackingGraph.get_nodes_with_attribute`.

I was going to separate out the graph validation in the `__init__` and add a flag as discussed in #91 but it wasn't a straightforward change (the validation is intertwined with the construction of the nodes_by_flag dict and we don't check edge validity at all, which we might want to add) so I will leave it for a future PR.

src/traccuracy/_tracking_graph.py
```diff
@@ -267,172 +254,51 @@ def get_location(self, node_id):
...
-            (List(hashable)): A list of node_ids which have the given attribute
+            (List(hashable)): A list of node_ids which have the given flag
                 and the value is True.
         """
-        if not isinstance(attr, NodeAttr):
-            raise ValueError(f"Function takes NodeAttr arguments, not {type(attr)}.")
-        return list(self.nodes_by_flag[attr])
+        if not isinstance(flag, NodeFlag):
+            raise ValueError(f"Function takes NodeFlag arguments, not {type(flag)}.")
+        return list(self.nodes_by_flag[flag])
```
msschwartz21, line 269: I remember we had a conversation about casting the set to a list before returning here, but I can't remember why we thought it was necessary. What was the logic?

src/traccuracy/track_errors/divisions.py
```diff
@@ -6,7 +6,7 @@
 from collections import Counter
 from typing import TYPE_CHECKING
 
-from traccuracy._tracking_graph import NodeAttr
+from traccuracy._tracking_graph import NodeFlag
```
msschwartz21, line 9: This comment applies broadly to this file. If we want to enforce the pattern of avoiding ragged attributes on our graphs, then we need to add a few lines here to set all flags to a default and then flip them as needed. This would match the pattern in the ctc functions.

Thoughts on including this change in this PR vs deferring to a separate one?

tests/test_tracking_graph.py
```diff
@@ -244,7 +194,7 @@ def test_get_connected_components(complex_graph, nx_comp1, nx_comp2):
     assert track2.graph.edges == nx_comp2.edges
 
 
-def test_get_and_set_node_attributes(simple_graph):
+def test_get_and_set_flag_on_node(simple_graph):
```
msschwartz21, line 197:
```suggestion
def test_set_flag_on_node(simple_graph):
```
We got rid of the get functionality right? So this test could be renamed. 

I think there are a couple other tests that could be renamed similarly but I can't comment on all of them directly.

msschwartz21 (changes requested):
I'm liking how these changes came together. I had a couple questions and potential places where tests could be renamed but they are all minor.

On a separate note, I want to put down a comment in writing in this thread so we have it if we ever look back. We are keeping the `set_flag_on_node/edge` and `set_flag_on_all_nodes/edges` functions (even though they are easily accessible through the networkx api) because it enables us to maintain dictionaries of nodes/edges by flag for faster lookup.

src/traccuracy/track_errors/divisions.py
DragaDoncila: @msschwartz21 perhaps should be in a separate PR? I'm wondering where you think it's best to first set them?

We could set them on graph creation, such that they are available and default `False` until the metrics are computed.

We could set them when the user requests a metrics computation, that way their mere presence indicates the graph has had metrics computed?

I kind of like having them on the graph from the beginning, but then there's nothing that can quickly tell you whether or not a graph has had its errors annotated...

tests/test_tracking_graph.py
```diff
@@ -254,26 +204,44 @@ def test_get_and_set_node_attributes(simple_graph):
...
+        NodeFlag.FALSE_POS: False,
     }
+
+    simple_graph.set_flag_on_all_nodes(NodeFlag.FALSE_POS, value=False)
+    for node in simple_graph.nodes:
+        assert simple_graph.nodes[node][NodeFlag.FALSE_POS] is False
+
+    simple_graph.set_flag_on_all_nodes(NodeFlag.FALSE_POS, value=True)
+    for node in simple_graph.nodes:
+        assert simple_graph.nodes[node][NodeFlag.FALSE_POS] is True
```
DragaDoncila, line 222: would be good to check also the `node_by_flag` dictionary here, and same below for edges, to ensure it's updated correctly

DragaDoncila (commented):
@cmalinmayor looking good! Left some comments through the code.

src/traccuracy/track_errors/divisions.py
msschwartz21: Given the need for discussion, let's leave this for another PR.

I don't want to put these flags on the graph at initialization because it will create more interdependency between the loaders and the errors. Right now someone adding a new metric/error shouldn't need to touch the loaders or `TrackingGraph`.

I think it's easier to slot them in at the beginning of metric calculation. We also added flags to `TrackingGraph` that record if node/edge/division errors have been annotated.

src/traccuracy/_tracking_graph.py
cmalinmayor: @msschwartz21 I agree that I don't remembery why we casted to a list. I tried it as a set and all the tests pass. @DragaDoncila  I looked at the networkx code to see if using `NodeView` or `EdgeView` makes sense, and even within networkx they only use those specific classes for `.nodes` and `.edges`. Other functions on the graph either return an iterable of ids (e.g. `Graph.neighbors()` [here](https://github.com/networkx/networkx/blob/862b0e95d2008c7e032fbdb907168927c24b4255/networkx/classes/graph.py#L1315)) or a different, custom dict-like view (e.g. `AdjacencyView` is returned from `DiGraph.succ()` [here](https://github.com/networkx/networkx/blob/862b0e95d2008c7e032fbdb907168927c24b4255/networkx/classes/digraph.py#L384), `DegreeView` from `Graph.degree()` [here](https://github.com/networkx/networkx/blob/862b0e95d2008c7e032fbdb907168927c24b4255/networkx/classes/graph.py#L1482)). This makes me want to keep it as a set or an generic iterable of ids for the `nodes_by_flag` case. If we end up needing the dictionary like access to attributes we can look into using an `AdjacencyView` or one of the other views, but I think it might be unnecessarily complex if an iterator will do and is consistent enough.

src/traccuracy/_tracking_graph.py
```diff
@@ -448,7 +314,7 @@ def get_preds(self, node):
         """
         return [pred for pred, _ in self.graph.in_edges(node)]
 
-    def get_succs(self, node):
+    def get_succs(self, node: Hashable) -> list[Hashable]:
```
cmalinmayor, line 317: We should revisit our re-implementation of get_preds and get_succs @msschwartz21. I think we didn't know we were exactly re-implementing [Graph.successors](https://github.com/networkx/networkx/blob/862b0e95d2008c7e032fbdb907168927c24b4255/networkx/classes/digraph.py#L874) since we use `out_edges` in our implementation - at the very least we should use the more appropriate networkx function. Is this still worth wrapping, so that we don't need to call `tracking_graph.graph.successors(n)`? Should we mimic the function name?

src/traccuracy/track_errors/divisions.py
cmalinmayor: I opened #129 for this discussion, and will mark this as resolved for this PR.

src/traccuracy/_tracking_graph.py
```diff
@@ -494,129 +360,133 @@ def get_subgraph(self, nodes):
...
+                f"Provided  flag {flag} is not of type NodeFlag. "
+                "Please use the enum instead of passing string values."
             )
-        for _id in ids:
-            self.graph.nodes[_id][attr] = value
-            if value:
-                self.nodes_by_flag[attr].add(_id)
-            else:
-                self.nodes_by_flag[attr].discard(_id)
+        self.graph.nodes[_id][flag] = value
```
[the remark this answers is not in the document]
cmalinmayor: Agreed - luckily we separated error annotation and metric computation, so new metrics that only use the existing annotations will be totally fine. New metrics probably shouldn't modify the values of existing NodeFlags or EdgeFlags. But any metric that adds a new Flag to the NodeFlag or EdgeFlag will have to be carefully checked.

src/traccuracy/_tracking_graph.py
msschwartz21: Good catch! Let's just drop the functions entirely. I think we really only use them in the division metrics and it doesn't add that much to just use the `networkx` api in this case.

src/traccuracy/_tracking_graph.py
DragaDoncila: @cmalinmayor thanks for looking at what `nx` is doing. I think since they just return iterables I think it's fine to keep as a set - I'd hate to cast for no reason.

cmalinmayor:
Closes #110 and #89

msschwartz21 (approved):
Docs look good! Code looks good! I'm happy :smile:

Issue #77, opened 2023-10-31
Documentation Update: Give each metric a page on the sidebar


cmalinmayor:
Each metric should have its own page on the sidebar with: a figure, an equation, a textual explanation, and any references. For this issue:
- [x] Set this format up within the RTD ecosystem with a dummy page for each implemented metric
- [ ] Create an example with a figure, an equation, a textual explanation, and references for one of the metrics
- [ ] Create issues for the other metrics (one each) (or do them)

PR #112, opened 2023-11-05
Adding track overlap metrics from `laptrack`

yfukai:
Adding the track purity (TP) and target effectiveness (TE) scores from `laptrack`. See https://github.com/yfukai/laptrack/issues/374 for discussion.

Definitions of the metrics are as follows. They are used in several literatures as the metrics that penalize the false negative and the false positive edges, respectively. 

- TE for a single ground truth track $T^g_j$ is calculated by finding the predicted track $T^p_k$ that overlaps with $T^g_j$ in the largest number of the frames and then dividing the overlap frame counts by the total frame counts for $T^g_j$. The TE for the total dataset is calculated as the mean of TEs for all ground truth tracks, weighted by the length of the tracks.

- TP is defined analogously, with $T^g_j$ and $T^p_k$ being swapped in the definition.


DragaDoncila:
Thanks for opening this @yfukai - very excited to get this in. I will review as soon as possible!

yfukai:
Thanks @DragaDoncila! The implementation may be sub-optimal, so I'm happy to have feedbacks on this!

src/traccuracy/metrics/_track_overlap.py
```diff
@@ -0,0 +1,108 @@
...
+            if mapped_nodes1 and mapped_nodes2:
+                for n1, n2 in product(mapped_nodes1, mapped_nodes2):
+                    reference_tracklet_edges_mapped.append((n1, n2))
+        overlaps = [
+            len(set(reference_tracklet_edges_mapped) & set(overlap_tracklet.edges()))
+            for overlap_tracklet in overlap_tracklets
+        ]
+        max_overlap = max(overlaps)
+        correct_count += max_overlap
+        total_count += len(reference_tracklet_edges_mapped)
```
cmalinmayor, line 106: Is there a reason the total is computed using only the mapped edges instead of `len(reference_tracklet)`? Based on my understanding of the metric, we want to divide by the total length of the reference tracklet, which is different from the mapped edges if there is a false negative node (the `if` on line 97 is false).

cmalinmayor (commented):
@yfukai Thank you so much for implementing these metrics! I just took a quick look and the code looks great - easy to follow, well commented, sensible changes to the existing code, has tests! I'll need to look more closely at the efficiency of the implementation because I do have a feeling that there might be a faster way, and I had one question about the correctness, but otherwise it looks fantastic. 

For better or worse, we are doing a pretty active sprint on the library this week, so we will probably introduce some breaking changes, but if we can get this merged with tests before, it will be on us to fix the issues we introduce 😅. Also, if you have any suggestions for improving the overall structure of the library, please add issues!

I will probably follow up with further documentation requests - we are planning to add a page in the docs for each metric as discussed in #77. But I'm happy to merge this first and have a new PR for that documentation.

Closes #112

src/traccuracy/metrics/_track_overlap.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self, matched_data: "Matched", include_division_edges: bool = True):
+        """Calculate metrics for longest track overlaps.
+
+        - Target Effectiveness
+        - Track Purity
+
+        Args:
+            matched_data (Matched): Matched object for set of GT and Pred data
+            include_division_edges (bool, optional): If True, include edges at division.
+        """
```
msschwartz21, line 37: In order for the docs to render correctly, can you move this docstring under the class (line 26) itself instead of under init?

src/traccuracy/metrics/_track_overlap.py
```diff
@@ -0,0 +1,108 @@
...
+        self.include_division_edges = include_division_edges
+        super().__init__(matched_data)
+
+    def compute(self):
+        # requires tracklets that also have the splitting and merging edges
+        # edgess are a list of edges, grouped by the tracks
```
msschwartz21, line 43: What is this comment referring to? Should it be removed?

src/traccuracy/metrics/_track_overlap.py
cmalinmayor: It would be good to test more cases like this to ensure there aren't any other edge cases we are missing - false negative nodes, false positive nodes, etc.

msschwartz21 (commented):
Thank you for contributing! I left two minor comments below. Caroline is correct that some of our open PRs are going to cause all sorts of merge conflicts, but we'll help you get that sorted out.

src/traccuracy/metrics/_track_overlap.py
yfukai: Oh thanks for noticing this, this is the left-off from the copied code. I'll remove this comment!

src/traccuracy/metrics/_track_overlap.py
yfukai: I moved the __init__ docstring to under the class!

src/traccuracy/metrics/_track_overlap.py
yfukai: Thanks for your comment! Yeah right, I believe it was not correct that I mapped the reference tracklet. I reorganized the code so that it maps the overlapping tracklet. I also added test with false negative nodes, false positive nodes and overlapping detections. I also checked the values of TE and TP swaps with swapping gt and pred tracks.

yfukai:
Thanks for your review @cmalinmayor and @msschwartz21! I believe I could resolve those points. I'm happy with either of merging this now or waiting for API update and resolve conflicts.

src/traccuracy/metrics/_track_overlap.py
```diff
@@ -0,0 +1,119 @@
...
+    overlap_tracklets: List[TrackingGraph],
+    overlap_reference_mapping: List[Tuple[Any, Any]],
+):
+    """Calculate weighted sum of the length of the longest overlap tracklet
+    for each reference tracklet.
+
+    Args:
+       reference_tracklets (List[TrackingGraph]): The reference tracklets
+       overlap_tracklets (List[TrackingGraph]): The tracklets that overlap
+       mapping (List[Tuple[Any, Any]]): Mapping between the reference tracklet nodes
```
cmalinmayor, line 82: Can you update this docstring? Thanks!

src/traccuracy/metrics/_track_overlap.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def compute(self):
+        gt_tracklets = self.data.gt_graph.get_tracklets(
+            include_division_edges=self.include_division_edges
+        )
+        pred_tracklets = self.data.pred_graph.get_tracklets(
+            include_division_edges=self.include_division_edges
+        )
+
+        gt_pred_mapping = self.data.mapping
```
cmalinmayor, line 53: If you make these dictionaries, for example `gt_pred_mapping =  {gt_id: [pred_id, ...]}` and `pred_gt_mapping = {pred_id: [gt_id]}`, you can replace the map_node helper function in the _calc_overlap_score function with a simple dictionary lookup, and increase efficiency by not looping over all pairs to find matched edges.

cmalinmayor:
> Thanks for your review @cmalinmayor and @msschwartz21! I believe I could resolve those points. If this looks OK (maybe after checking the implementation?), I'm happy with either of merging this now or waiting for API update and resolving conflicts.

Thanks for the changes! I had two quick comments, one about a docstring and the other a suggestion to improve efficiency. After we resolve those I think we can merge now and worry about API changes later. Thanks!

src/traccuracy/metrics/_track_overlap.py
yfukai: done!

src/traccuracy/metrics/_track_overlap.py
yfukai: This is also done! I hope the implementation has become more efficient with this.

cmalinmayor (approved):
Thanks @yfukai for being so responsive! I'm happy to merge this now, unless @msschwartz21 has any final notes. Whoever updates this for the new API (probably Morgan or I) will request your review to double check any changes, and we will also be following up with a request for documentation soon. Super excited to have these metrics in the library!

msschwartz21 (approved):

msschwartz21 (approved):
LGTM! Thank you!

PR #116, opened 2023-11-10
Add written documentation for errors and metrics

msschwartz21:
I'm opening this as a draft so that we can take a look at the RTD preview while we are working on it.

Pages to write:
- [x] Track errors / divisions
- [ ] Track errors / edges
- [ ] Track errors / nodes
- [ ] Metrics / divisions
- [x] Metrics / ctc 
- [ ] Metrics / track overlap


msschwartz21:
@cmalinmayor I took a stab at the division errors page. I just used super simple text based diagrams like we use in the docstrings for tests. We may want to upgrade to something fancier in the future, but I don't want the plotting to become a blocker to writing.

docs/source/track_errors/divisions.rst
```diff
@@ -0,0 +1,95 @@
+Division Errors
+===============
+Note: These are annotated on the parent nodes.
+
+True Positive
+-------------
+
+A true positive division is a division event in which the parent and both daughters match between the ground truth and predicted graphs. True positive divisions are annotated on both the ground truth and predicted graphs.
```
cmalinmayor, line 8: Can you say that they are annotated on the parent node specifically? That's what we do, right?

cmalinmayor:
Ooh I like it a lot!! I think if we should implement this for all metrics and then we can make it fancier later if we have time and inclination, but this is so much better than nothing.

msschwartz21:
@DragaDoncila Caroline and I started this branch to work towards closing #77 by adding pages in the sidebar for each error annotation type and downstream metrics. Can you take a stab at the ctc related pages which I think would be `docs/track_errors/nodes.rst`, `docs/track_errors/edges.rst`, and `docs/metrics/ctc.rst` when you have some time?

DragaDoncila:
@msschwartz21 would you like me to PR into this branch for these docs, or review this PR independently and make another for the CTC docs

msschwartz21:
> @msschwartz21 would you like me to PR into this branch for these docs, or review this PR independently and make another for the CTC docs

I'd prefer to get all of the docs changes in place before we merge into main so let's do a PR or direct push to this branch.

DragaDoncila:
@msschwartz21 I merged some additions to the CTC docs!

msschwartz21:
Since we are starting to get a few more eyes on traccuracy, I'm leaning towards a mindset of some incomplete documentation is better than no documentation at all. To that end, I added a note block to the pages we haven't written yet saying "Coming soon!"

@DragaDoncila @cmalinmayor Does this sound ok to you? I'll open up an issue to keep track of the remaining pages that we need to write.

cmalinmayor:
I very much agree. Some docs better than no docs!

cmalinmayor (approved):

PR #117, opened 2023-11-13
Add `TrackOverlapMetrics` to benchmarking 

msschwartz21:


tests/bench.py
```diff
@@ -130,3 +130,10 @@ def run_compute():
     assert div_results["Frame Buffer 0"]["False Negative Divisions"] == 25
     assert div_results["Frame Buffer 0"]["False Positive Divisions"] == 31
     assert div_results["Frame Buffer 0"]["True Positive Divisions"] == 69
+
+
+def test_track_overlap_metrics(benchmark, iou_matched):
+    def run_compute():
+        return TrackOverlapMetrics(copy.deepcopy(iou_matched)).compute()
```
cmalinmayor, line 137: Do we still need these `deepcopy` calls with the new Matched objects? Yes, right, because we are reusing the Matched object across benchmarks and don't want the order that they run to affect the benchmark?

cmalinmayor:
Did the benchmarking fail just becuase it had an extra thing to run that wasn't in the baseline? How does the action deal with adding new benchmark functions?

tests/bench.py
msschwartz21: That was the original logic. Since it's a fixture, I was paranoid about cross-contaminating across the different benchmarking functions.

msschwartz21:
> Did the benchmarking fail just becuase it had an extra thing to run that wasn't in the baseline? How does the action deal with adding new benchmark functions?

Honestly I have no idea what's happening. The updated benchmarking runs fine locally for me. I don't think the addition of a new function should cause a problem. We manually make a comparison when we build a pandas dataframe in the "generate report" section. In that case, I think we should just end up with a missing value for the new function on the base commit.

@tlambert03 Any guidance on troubleshooting action failures like this? https://github.com/Janelia-Trackathon-2023/traccuracy/actions/runs/6855566985/job/18642828648?pr=117

tlambert03:
huh, strange.  hard to say what's canceling it huh?  i would just do some brute force stuff.  maybe add `-v` to the pytest call to get a bit more information on whether it's hanging on exactly the test you've added.  exclude/comment that test again just to ensure that it is indeed that test.  then dig into the test itself and perhaps put debugging statements to see where in the metric it's canceling?

msschwartz21:
Closing because this work is being addressed in #218

PR #105, opened 2023-11-03
Updating notebook for more elaborate tutorial 📔 

KristinaUlicna:
### PR summary:

Changes to the `ctc.ipynb` notebook to:
- [x] Detection of certificate error - closes #84 
- [x] Instruct the user on how to manually resolve ⬆️ - closes #102 
- [x] Instantiate the `TrackingGraph` upon each computation to avoid maintenance of previously computed attributes with `tracking_graph_data_loader` mini function hack - resolved #101 
- [x] Compute the metrics with moving IoU threshold to monitor the metrics / performance dependence ▶️ visualised on a plot 🎨 Closes #106

### What should the reviewer focus their feedback on?

- Overall sanity check (motivation behind the plot)
- Code documentation / annotation quality

PR #118, opened 2023-11-15
Update `run_metrics` to take instantiated `Matchers` and `Metrics` as input

### LINKED ISSUE #87: API Update: Instantiate Matcher and Metrics classes before passing into run_metrics

msschwartz21:
- Closes #87 
- I also eliminated an unused/unimplemented validation function. I still think we should implement some type of validation between metrics/matchers but it may be more complicated than this one function.

We will also need to update the ctc example notebook after this change. I can either make the updates in this PR or we can handle it in #105. @KristinaUlicna do you have a preference?


tests/test_run_metrics.py
```diff
@@ -0,0 +1,16 @@
+from traccuracy import run_metrics
+from traccuracy.matchers import CTCMatcher
+from traccuracy.metrics import CTCMetrics
+
+from tests.test_utils import get_movie_with_graph
+
+
+def test_run_metrics():
```
cmalinmayor, line 8: This test is minimal indeed. That said, what do we actually want to test in run_metrics that is not tested in the actual Metrics/Matchers? The only thing I can really thing of is that it works for multiple metrics, and then maybe validation when we implement it properly (although we will probably test the validation function separately, right?).

cmalinmayor (changes requested):
This change is SO MUCH NICER. I love it. My only hold-up is that we should avoid creating testing code debt - let's make a thorough run_metrics test, whatever that means. The new run_metrics is so minimal in terms of funcitonality that it might end up being very short and similar to what we have here, but let's think about it and see if we need more.

src/traccuracy/_run_metrics.py
```diff
@@ -13,43 +11,28 @@
...
-        m_class: get_relevant_kwargs(m_class, metrics_kwargs) for m_class in metrics
-    }
+    matched = matcher.compute_mapping(gt_data, pred_data)
     results = {}
     for _metric in metrics:
-        relevant_kwargs = metric_kwarg_dict[_metric]
-        result = _metric(**relevant_kwargs).compute(matched)
-        results[_metric.__name__] = result
+        result = _metric.compute(matched)
+        results[_metric.__class__.__name__] = result
```
cmalinmayor, line 37: What if someone runs with the same metric but different hyperparameters? That will probably happen with the new forumlation, even though it was impossible with the old one. Do we want to support this or not?

msschwartz21: This is a good catch. I think we should support it but I'll have to think about the right naming conventions to make it work.

tests/test_run_metrics.py
msschwartz21: I think I should add a case where we actually have more than one metric. I'm also thinking about switching to DummyMetric and DummyMatcher so we don't waste any time actually running the algorithms even on a simple test case. 

I had thought about trying to do more validation of the inputs which would add some extra cases for test coverage.

cmalinmayor: DummyMetric and DummyMatcher make sense to me, although having DummyMetric return a non-empty dictionary would be good just for sanity checking the results output format.

We can add tests for validation as we implement it - but it's good to have the space to put it now.

src/traccuracy/_run_metrics.py
cmalinmayor: Perhaps a str function on the metric classes that does something nice?

cmalinmayor: And/or the Metrics object optionally takes a name input?

msschwartz21: Let me know what you think of what I implemented. It should generate an output like this:
```
[
{'MetricName': results,
'parameters': {kwargs}},
{'MetricName': results,
'parameters': {kwargs2}}
]
```

src/traccuracy/_run_metrics.py
```diff
@@ -1,55 +1,48 @@
 from typing import TYPE_CHECKING
 
-from traccuracy._utils import get_relevant_kwargs, validate_matched_data
+# from traccuracy import TrackingGraph
+# from traccuracy.matchers._base import Matcher
+# from traccuracy.metrics._base import Metric
```
msschwartz21, line 5: I am trying to import these objects so that I can validate the input to `run_metrics` in lines 33-40; however, these imports are giving me circular import errors. @tlambert03 do you see a way to fix this and make the imports work?

tlambert03: two options:

one is just take the easy route and put those imports inside the `run_metrics` function itself (no shame in that).

there is a second option, not sure it's better... 
If you look at the stack trace:
```
tests/track_errors/test_divisions.py:4: in <module>
    from traccuracy import NodeAttr, TrackingGraph
src/traccuracy/__init__.py:9: in <module>
    from ._run_metrics import run_metrics
src/traccuracy/_run_metrics.py:3: in <module>
    from traccuracy import TrackingGraph
E   ImportError: cannot import name 'TrackingGraph' from partially initialized module 'traccuracy' (most likely due to a circular import) (/Users/talley/
```

you can see the order of things:
1. the test is trying to import `TrackingGraph` from `tracurracy.__init__`
2. while importing `tracurracy.__init__` it *first* imports `._run_metrics`
3. `_run_metrics` also tries to import `TrackingGraph` from `tracurracy.__init__`
4. we're back to trying to import `tracurracy.__init__` which still doesn't have `TrackingGraph` ready, and we have a problem

so, solution number two is to just move the import of TrackingGraph in the `__init__` file above the import of `TrackingGraph`... (and add a comment to tell ruff not to freak out about it):

```python
from ._tracking_graph import EdgeAttr, NodeAttr, TrackingGraph

# must go after TrackingGraph
from ._run_metrics import run_metrics  # isort:skip
```

not sure which is better... you can pick :)

src/traccuracy/_run_metrics.py
cmalinmayor: How about: 
```
[
{"results": results dict
"metric": metric dict including "name" entry},
]
```
This is temporary until we implement #120 but is closer to the desired access pattern.

src/traccuracy/_run_metrics.py
```diff
@@ -1,55 +1,51 @@
...
-            matcher class
-        metric_kwargs (optional, dict): Dictionary of any keyword args for the
-            Metric classes
+        gt_data (traccuracy.TrackingGraph): ground truth graph and optionally segmentation
+        pred_data (traccuracy.TrackingGraph): predicted graph and optionally segmentation
+        matcher (Matcher): instantiated matcher object
+        metrics (List[Metric]): list of instantiated metrics objects to compute
 
     Returns:
         Dict: dictionary of metrics indexed by metric name. Dictionary will be
```
cmalinmayor, line 30: This docstring is not updated to the new List structure

cmalinmayor (approved):
Looks great! Other than the one docstring I noted, I'm happy to merge - I'll just approve now so we don't need to go back and forth again. Thanks for implementing the tests! 🎉

cmalinmayor:
@msschwartz21 Actually one more issue - is the ReadTheDocs circular imports error the same one we already dealt with, or did we introduce a new one?

PR #119, opened 2023-11-17
Add CTC format checks in loader

bentaculum:
This adds some checks to enforce proper Cell Tracking Challenge data format already while loading, and prints descriptive error messages.
As discussed in the last meeting, they are turned on by default, but can be turned off with a boolean in `load_ctc_data`.

I've been dealing with tracking results in faulty CTC format recently and ran into weird errors downstream of the loader in the metrics computation, which inspired this PR.

The proposed routine checks:
- unique positive integer track IDs
- indicated parent IDs present
- parent tracklet ends before child tracklet starts (gaps possible)
- masks contain all detections as indicated in tracks file, and vice versa.

### DIFF
--- a/src/traccuracy/loaders/_ctc.py
+++ b/src/traccuracy/loaders/_ctc.py
@@ -5,7 +5,7 @@ import os
 import networkx as nx
 import numpy as np
 import pandas as pd
-from skimage.measure import regionprops_table
+from skimage.measure import label, regionprops_table
 from tifffile import imread
 from tqdm import tqdm
 
@@ -157,23 +157,26 @@ def ctc_to_graph(df, detections):
     return G
 
 
-def _check_ctc(tracks: pd.DataFrame, detections: pd.DataFrame):
+def _check_ctc(tracks: pd.DataFrame, detections: pd.DataFrame, masks: np.ndarray):
     """Sanity checks for valid CTC format.
 
-    Checks for:
+    Hard checks (throws exception):
     - Tracklet IDs in tracks file must be unique and positive
     - Parent tracklet IDs must exist in the tracks file
     - Intertracklet edges must be directed forward in time.
     - In each time point, the set of segmentation IDs present in the detections must equal the set
     of tracklet IDs in the tracks file that overlap this time point.
 
+    Soft checks (prints warning):
+    - No duplicate tracklet IDs (non-connected pixels with same ID) in a single timepoint.
 
     Args:
         tracks (pd.DataFrame): Tracks in CTC format with columns Cell_ID, Start, End, Parent_ID.
         detections (pd.DataFrame): Detections extracted from masks, containing columns
             segmentation_id, t.
+        masks (np.ndarray): Set of masks with time in the first axis.
     Raises:
-        ValueError: If any of the checks fail.
+        ValueError: If any of the hard checks fail.
     """
     logger.info("Running CTC format checks")
     if tracks["Cell_ID"].min() < 1:
@@ -206,6 +209,12 @@ def _check_ctc(tracks: pd.DataFrame, detections: pd.DataFrame):
             raise ValueError(
                 f"IDs {det_ids - track_ids} at t={t} not represented in tracks file."
             )
+
+    for t, frame in enumerate(masks):
+        _, n_components = label(frame, return_num=True)
+        n_labels = len(detections[detections["t"] == t])
+        if n_labels < n_components:
+            logger.warning(f"{n_components - n_labels} non-connected masks at t={t}.")
 
 
 def load_ctc_data(data_dir, track_path=None, run_checks=True):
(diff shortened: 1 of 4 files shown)


msschwartz21:
I haven't looked at the changes yet, but just wanted to let you know that the docs failure is my bad. I think I have it fixed on some open PR but I'm not sure which one. Is it ok if I push a commit to this branch fixing the docs issue?

bentaculum:
No worries, and yes of course go ahead.

src/traccuracy/loaders/_ctc.py
```diff
@@ -157,13 +160,54 @@ def ctc_to_graph(df, detections):
...
+        track_ids = set(
+            tracks[(tracks["Start"] <= t) & (tracks["End"] >= t)]["Cell_ID"]
+        )
+        det_ids = set(detections[(detections["t"] == t)]["segmentation_id"])
+        if not track_ids.issubset(det_ids):
+            raise ValueError(f"Missing IDs in masks at t={t}: {track_ids - det_ids}")
+        if not det_ids.issubset(track_ids):
+            raise ValueError(
+                f"IDs {det_ids - track_ids} at t={t} not represented in tracks file."
+            )
```
msschwartz21, line 200: Just to double check my understanding, we expect to `track_ids == det_ids`, but by testing it this way we can give a more informative error message?

msschwartz21 (commented):
High level question/thought. The way this function is currently written it will fail on the first detected error. So if someone is trying to debug issues with their data they will repeatedly have to run this function to catch each error individually. Alternatively, we could collect the messages from `ValueError` in a list of errors and then raise one final error at the end if anything is in the list. This way we would produce a complete report of all issues in one go. What do you think based on the use case you've encountered?

cmalinmayor (changes requested):
Thanks @bentaculum, this is great! I like Morgan's idea of making it run all the checks even if one fails to provide all types of failure information without having to re-run, but even as-is I can see this being very useful. Can you add to the docstring what it actually checks for and that it raises a ValueError? To test my understanding and help you out with my request, I'll draft what it checks for here 🙂 
```
Checks for:
- Tracklet IDs in tracks file must be unique and positive
- Parent tracklet IDs must exist in the tracks file
- Intertrack edges must go forward in time (the last cell in the parent tracklet has to be earlier than the first cell in the child tracklet)
- In each time point, the set of segmentation IDs present in the detections must equal the set of tracklet IDs in the tracks file (Question: in practice, does this mean there is a one-to-one correspondence? I think the way the tracks and detections are represented the answer is yes, because neither allows duplicate IDs in one time frame?).
```

bentaculum:
> Thanks @bentaculum, this is great! I like Morgan's idea of making it run all the checks even if one fails to provide all types of failure information without having to re-run, but even as-is I can see this being very useful. 

The errors are usually very few edge cases that somehow went wrong in some data format conversion, so I think simply aborting with a single descriptive error message when there's an error in the format is sufficient.

cmalinmayor (approved):
Look great! Thanks @bentaculum 🎉

bentaculum:
I just added one more check for non-connected pixels with the same ID (=more than one connected component for a label ID) in a single frame. This unfortunately does occur in some of the CTC datasets, here's a screenshot from `Fluo-N2DL-HeLa/train/01`, so I added it as a soft check that prints a warning.
Screenshot 2023-12-05 at 18 37 08

cmalinmayor:
One question re: benchmarking. Should we set the run_checks flag to True or False in the benchmarking? Or perhaps set it to False and benchmark the checking separately? @bentaculum @msschwartz21 

After the most recent mask-based check there seems to be a 30%ish increase in the loading benchmarks, I assume with the default True value compared to not checking, which is actually super useful to know.

msschwartz21:
> One question re: benchmarking. Should we set the run_checks flag to True or False in the benchmarking? Or perhaps set it to False and benchmark the checking separately?

Good catch. Let's pull it out as a separate function and set the flag to False when we load data elsewhere in benchmarking. This should include the fixtures for loading data.

bentaculum:
> > One question re: benchmarking. Should we set the run_checks flag to True or False in the benchmarking? Or perhaps set it to False and benchmark the checking separately?
> 
> Good catch. Let's pull it out as a separate function and set the flag to False when we load data elsewhere in benchmarking. This should include the fixtures for loading data.

Nice idea, just did as you suggested, comparative benchmark to `main` looks good now https://github.com/Janelia-Trackathon-2023/traccuracy/commit/c15c2f654a213fbc52277ef236a4fc765f1973cf#comments

I guess the comparison to the new benchmark will appear once it is merged to main.

msschwartz21 (approved):
Thanks for making that last change! I think you're good to merge! 

As far as I know, the new benchmark diff should show up once this is merged into main 🤞 , but this is the first time we're adding a new benchmark function.

Issue #83, opened 2023-10-31
Feature: Validation of Matchers and Metrics that work together


cmalinmayor:
As we expand beyond CTC metrics, we know that certain matchers and metrics do not make sense to be run together because they make different assumptions. We want to formalize the assumptions that each matcher and metric makes, and perform some validation based on that formalization. 

Some example assumptions we have discussed include:

- one-to-one matching or one-to-many matching or many-to-one matching - does the matching allow multiple ground truth nodes to be matched to one predicted node, or vice versa? 
- matching on nodes vs matching on edges - this is potentially a subset of the above assumptions, but should be considered
- dense ground truth
- has segmentation mask

This validation could be done in run_metrics, or elsewhere. We need to pin down the concepts and then determine where/how in the code to put the validation (and then document!).

msschwartz21:
Just want to drop a thought here after Caroline and I were discussing this issue. 

Given that we have a very limited number of matchers and metrics, it might be easier to just hardcode a list of compatible matchers/metrics on each matcher/metric object. The we can just be thoughtful about what is compatible and not have to come up with a super clever system for determining it on an automated basis.

PR #124, opened 2023-11-18
Fix subclass docstrings for sphinx autoapi

### LINKED ISSUE #72: Sphinx autoapi documentation of subclasses shows the docstring of the base class
We use abstract base classes in a couple places in the repo. The subclasses are fully documented with docstrings, but when sphinx build the documentation it ends up using the docstring from `ABC` instead of the docstring that we wrote. See [example](https://traccuracy.readthedocs.io/en/latest/autoapi/traccuracy/matchers/index.html#traccuracy.matchers.CTCMatched). 

@tlambert03 have you come across this issue before? I've looked into it a couple times and have failed to fix it every time. At this point, I'm so frustrated by it that I would be willing to switch to an entirely different system for building the docs if it made the issue go away.

msschwartz21:
# Proposed Change
Fixes outstanding issues with docstrings to close #72 

# Types of Changes
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] Documentation update
- [ ] Tests and benchmarks

- [ ] Loaders
- [ ] Matchers
- [ ] Track Errors
- [ ] Metrics

### DIFF
--- a/src/traccuracy/metrics/_divisions.py
+++ b/src/traccuracy/metrics/_divisions.py
@@ -70,8 +70,8 @@ class DivisionMetrics(Metric):
         """Runs `_evaluate_division_events` and calculates summary metrics for each frame buffer
 
         Args:
-        matched_data (traccuracy.matchers.Matched): Matched object for set of GT and Pred data
-            Must meet the `needs_one_to_one` criteria
+            matched_data (traccuracy.matchers.Matched): Matched object for set of GT and Pred data
+                Must meet the `needs_one_to_one` criteria
 
         Returns:
             dict: Returns a nested dictionary with one dictionary per frame buffer value
--- a/src/traccuracy/metrics/_divisions.py
+++ b/src/traccuracy/metrics/_divisions.py
@@ -57,8 +57,6 @@ class DivisionMetrics(Metric):
         3, 734559 (2021).
 
     Args:
-        matched_data (Matched): Matched object for set of GT and Pred data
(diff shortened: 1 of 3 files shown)


src/traccuracy/metrics/_divisions.py
```diff
@@ -71,6 +69,10 @@ def __init__(self, frame_buffer=(0,)):
     def compute(self, data: Matched):
         """Runs `_evaluate_division_events` and calculates summary metrics for each frame buffer
 
+        Args:
+            matched_data (traccuracy.matchers.Matched): Matched object for set of GT and Pred data
+                Must meet the `needs_one_to_one` criteria
```
cmalinmayor, line 74: TBH I still don't know what this means. We are saying that the division methods only work if the matching is one-to-one? I'd expect the metric to have the `needs_one_to_one` criteria and the matched to have... `one_to_one`? Saying that the Matched needs one-to-one doesn't make sense linguistically. But, we can relegate this to our validation issue #83

cmalinmayor (approved):
You can choose to address my problem/confusion with the one-to-one docstring or leave it for #83, I'm fine either way. Thanks!

src/traccuracy/metrics/_divisions.py
msschwartz21: Yeah I'm going to kick the can down the road for this one and leave it for #83. I agree that it's not clear, but I don't think there is an easy quick fix.

PR #125, opened 2023-11-20
fix import loop

tlambert03:
this might be a better way to avoid circular imports, and appears to work for docs too

### DIFF
--- a/src/traccuracy/__init__.py
+++ b/src/traccuracy/__init__.py
@@ -6,9 +6,7 @@ try:
 except PackageNotFoundError:  # pragma: no cover
     __version__ = "uninstalled"
 
+from ._run_metrics import run_metrics
 from ._tracking_graph import EdgeAttr, NodeAttr, TrackingGraph
-
-# must go after TrackingGraph to avoid circular imports
-from ._run_metrics import run_metrics  # isort:skip
 
 __all__ = ["TrackingGraph", "run_metrics", "NodeAttr", "EdgeAttr"]
--- a/src/traccuracy/_run_metrics.py
+++ b/src/traccuracy/_run_metrics.py
@@ -1,6 +1,6 @@
 from typing import TYPE_CHECKING
 
-from traccuracy import TrackingGraph
+from traccuracy._tracking_graph import TrackingGraph
 from traccuracy.matchers._base import Matcher
 from traccuracy.metrics._base import Metric
 
--- a/src/traccuracy/track_errors/_ctc.py
+++ b/src/traccuracy/track_errors/_ctc.py
@@ -4,7 +4,7 @@ from typing import TYPE_CHECKING
 
 from tqdm import tqdm
 
-from traccuracy import EdgeAttr, NodeAttr
+from traccuracy._tracking_graph import EdgeAttr, NodeAttr
 
 if TYPE_CHECKING:
     from traccuracy.matchers import Matched


msschwartz21 (approved):
Thank you, Talley! Just so it's in writing, our policy should be to avoid importing directly from `traccuracy` within the library even if we have an import exposed there for convenience.

tlambert03:
>  our policy should be to avoid importing directly from traccuracy within the library even if we have an import exposed there for convenience.

yeah, I guess at least as a way to resolve circular imports.   In other words, my two suggestions previously were both less than ideal, and we just needed to go down the rabbit hole a bit further.  You might sometimes be able to get away with the smaller higher level imports in internal usage... but it's definitely the first thing to check if not

DragaDoncila:
@tlambert03 just a general question - why is it so common for projects to run into circular import issues? Presumably if this was a design problem with Python itself it would've been solved already somehow, since it's pretty obvious that people will want to import parts of their package into other parts of their package. So is the fact that we're running into circular imports indicative of poor code design? What's the "blessed" way that a perfect project would handle this? And if there is no such blessed path, what's the crux of the issue?

tlambert03:
yeah, I don't think it can be considered a design problem with Python itself (unless we're prepare to say the python being an interpreted language is itself a design flaw :))

> since it's pretty obvious that people will want to import parts of their package into other parts of their package

yes, but the details matter here, and I'm not sure it's *as* common that module A needs to import module B, and module B needs to import module A (for reasons beyond just type annotations).  If that's truly the case, then I think you probably should make those a single module.

> So is the fact that we're running into circular imports indicative of poor code design?

I think it's something closer to this, but probably less so than you might think?

>  What's the "blessed" way that a perfect project would handle this? And if there is no such blessed path, what's the crux of the issue? Why is it that most packages I seem to work with at some stage run into circular imports?

Here's my stab at the crux of the issue (some of these points are obvious)
1. naturally, people like to use modules to break up their code into manageable bits (it's worth remembering that we would never have circular imports if we didn't use modules :smile:)
2. it's considered best practice to put all of your imports at the top of the module (people tend to dislike putting imports inside local scopes)
3. people tend to use things like isort/ruff to sort their imports more or less alphabetically
4. at runtime, python interprets the file line by line (meaning import order *does* matter)
5. some packages like to make all of their stuff available at the top level of their package (that's also fine and good, certainly makes for an easier public API)
6. **crux** 👉: if you also *consume* your own API using those pretty top level exports, then you've got a recipe for circular imports.

here's a simplified example of what you're running into here, as a result of both exporting and consuming names from the top level package:

```
my_package/
├── __init__.py
├── mod_a.py
└── mod_b.py
```

```python
# my_package/__init__.py
from .mod_a import ThingA
from .mod_b import ThingB
__all__ = ["ThingA", "ThingB"]
```

```python
# my_package/mod_a.py
from my_package import ThingB  # 👀 using internal thing from top level

class ThingA:
    B = ThingB()
```

```python
# my_package/mod_b.py
class ThingB:
    ...
```

now try to import:

```sh
$ python -c "import my_package"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/talley/dev/forks/traccuracy/src/my_package/__init__.py", line 1, in <module>
    from .mod_a import ThingA
  File "/Users/talley/dev/forks/traccuracy/src/my_package/mod_a.py", line 1, in <module>
    from my_package import ThingB
ImportError: cannot import name 'ThingB' from partially initialized module 'my_package'
(most likely due to a circular import) (/Users/talley/dev/forks/traccuracy/src/my_package/__init__.py)
```

the stack trace shows you the crux of it: as python goes line by line through `__init__.py`, *while* you're in the process of importing ThingA from `.mod_a`, `mod_a` itself tries to import something else that is lower down in the `__init__.py` file.

```mermaid
graph TD
    init[("__init__.py")]
    modA[("mod_a.py")]
    modB[("mod_b.py")]

    init -->|ThingA| modA
    init -->|ThingB| modB
    modA -->|ThingB| init
```

> **NOTE:** I intentionally set that up though... if we switched it around and  put `from my_package import ThingA` inside of `mod_b`, then it would have "accidentally" worked... just because alphabetically `mod_b` would have come second, and `ThingA` would already be in the namespace of the top level `__init__` file by the time `mod_b` was imported.


**TLDR;**, if you use modules, and you don't want to put your imports in local scopes, and you don't want the (alphabetical) order of imports to be critically important to avoid circular imports, then don't consume your own top level exports.  Import internally from lower down:

```python
# my_package/mod_a.py
from my_package.mod_b import ThingB  # 👀 using internal thing from mod_b, not top level

class ThingA:
    B = ThingB()
```

and we no longer have a cycle:

```mermaid
graph TD
    init[("__init__.py")]
    modA[("mod_a.py")]
    modB[("mod_b.py")]

    init -->|ThingA| modA
    modA -->|ThingB| modB
    init -->|ThingB| modB
```

DragaDoncila:
@tlambert03 thanks for the detailed response. I think it's the "consuming our own API" bit that made it click for me. Even if something is exported at the top level for people to use, I think it makes sense that internally, we import it from its original location. And then we'll never run into circular imports again 😀

PR #127, opened 2023-12-04
Add tests for correct usage of gap-closing or frame-skip edges

### LINKED ISSUE #115: Feature request: supporting edges connecting non-sequential frames
* traccuracy version: 0.0.2
* Python version: any 
* Operating System: any

### Description

Currently the edges of the tracking graph are defined as follows:
> Edges point from a node representing a cell in time point t to the same cell or its daughter in t+1.

However, I think some tracking models (including the popular LAP tracker implemented in the `TrackMate`, `utrack` and `laptrack`) can detect edges connecting nodes at non-sequential frames by assuming some cells are undetected in the skipped frames. To keep the metrics fair to those algorithms, it may be useful to revise the code so that they can deal with the frame-skipping edges. Thanks!

### What I Did

not applicable

DragaDoncila:
# Proposed Change
This PR adds tests to ensure gap closing edges are loaded and evaluated correctly by existing metrics and matchers, and updates language describing graphs to include definitions and discussion of such edges.

# Types of Changes
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] Documentation update
- [x] Tests and benchmarks

- [x] Loaders
- [x] Matchers
- [x] Track Errors
- [ ] Metrics

# Checklist

# Further Comments
Closes #115.


tests/track_errors/test_divisions.py
```diff
@@ -220,3 +220,30 @@ def test_evaluate_division_events():
...
+    assert g_gt.nodes["1_1"][NodeAttr.FN_DIV]
+
+    # fix division, assert it's identified correctly
+    g_pred.remove_node("2_2")
+    g_pred.add_edge("1_1", "2_3")
+    # mapper doesn't need to change as removed node was always missing
+    matched_data = Matched(TrackingGraph(g_gt), TrackingGraph(g_pred), mapper)
+    _classify_divisions(matched_data)
+    assert g_gt.nodes["1_1"][NodeAttr.TP_DIV]
+    assert g_gt.nodes["1_1"][NodeAttr.TP_DIV]
```
msschwartz21, line 240:
```suggestion
    assert g_pred.nodes["1_1"][NodeAttr.TP_DIV]
```

tests/test_utils.py
```diff
@@ -148,3 +148,124 @@ def get_division_graphs():
...
+def get_gap_close_graphs():
+    """
+    G1
+                                      3_5 -- 3_6 -- -- -- 5_10
+    1_0 -- 1_1 -- -- -- 2_3 -- 2_4 -<
+                                      4_5 -- 4_6
+    G2
+                                      2_5 -- 2_6 -- -- -- 4_10
+    1_0 -- 1_1 -- 1_2 -- 1_3 -- 1_4 -<
+                                      3_5 -- 3_6
```
msschwartz21, line 162: @cmalinmayor I thought we had validation in the `TrackingGraph` about edges only spanning one frame, but these graphs don't seem to be causing errors and I can't find this validation in the `TrackingGraph`. Am I misremembering or did we drop it at some point?

msschwartz21 (commented):
This looks like a good set of initial test cases! Should each test case also be run in the opposite direction, e.g. switching gt and pred? 

For the docs, I think you made the appropriate updates in what is currently written. In #116, there will be some places to add more content about the impact of frame-skip edges. 

I expected we would need some changes in `TrackingGraph` and associated testing, but it looks like that isn't the case since everything is passing!

tests/test_utils.py
lupinthief: I was running into the `TrackingGraph` validation wanting 't' for every `node_id`, which stemmed from my data having gaps.

cmalinmayor: > @cmalinmayor I thought we had validation in the `TrackingGraph` about edges only spanning one frame, but these graphs don't seem to be causing errors and I can't find this validation in the `TrackingGraph`. Am I misremembering or did we drop it at some point?

@msschwartz21 I also thought this was in the validation but I think we actually didn't add it, perhaps on accident...

cmalinmayor: > I was running into the `TrackingGraph` validation wanting 't' for every `node_id`, which stemmed from my data having gaps.

@lupinthief In my current understanding of implementing tracks with gaps, we still want all nodes to have a time associated. The gaps in time are covered by the edges that skip over frames, but still connect nodes that have times (e.g. an edge from a node in time `t` to a node in time `t+n`). Can that cover your use case as well?

lupinthief:
Yes, that covers my use case, it is just that the edges were determined
from a monotonic range between 'start' and 'end', while the node attributes
were assigned from the 'detections_df' which didn't have a continuous set
of timesteps leading to empty nodes.

DragaDoncila:
@lupinthief thanks for taking a look at this and glad to hear it should work for your use case!

I'm still not sure I understand how exactly your data is set up, and I'm curious to learn more.

> it is just that the edges were determined
from a monotonic range between 'start' and 'end', while the node attributes
were assigned from the 'detections_df' which didn't have a continuous set
of timesteps leading to empty nodes.

What do `empty nodes` mean here? Are they nodes that had edges connecting them, but were for some reason not present in `detections_df` and therefore had no further attributes associated, including `t`? If so, what causes the discrepancy between `detections_df` and the nodes present in the graph?

lupinthief:
Thanks for the interest. I'm loading the lineage in LBEP format, and the
CTC to graph function (I forget precise name) basically created edges
assuming a continuous no.arange between 'beginning'  and 'end+1'. The
'detections_df' was missing some of those timesteps so I ended up with
exactly the situation you describe.  Sorry, I'm away from my machine so
working from memory here.

DragaDoncila:
@msschwartz21 @cmalinmayor I've resolved conflicts and updated the tests here. One thing I'm not sure about here is whether we want to do a test for divisions with frame tolerance, or whether those divisions get "shifted" before being counted. Feedback welcome on more tests I can add!

msschwartz21:
@DragaDoncila Thanks for getting this updated! When taking a look through this I realized we have made a lot of changes to our testing framework since this branch first got started. I started a branch to add the gap closing cases in to our standard testing examples and used those in the tests for ctc errors. I still need to write the division and basic error tests before it's ready to get merged in here. 

Re frame shifted divisions: Caroline and I started talking about it and I think we need to write the tests if only to understand what the behavior is and document it. I'm guessing we will essentially say we don't support gap edges for shifted divisions.

DragaDoncila:
@msschwartz21 yeah I meant to write a comment about that, but forgot. If you have partial code and are busy with other stuff I'm happy to update the tests to match the new framework. Otherwise, feel free to close this and work on a new branch if you'd like.

DragaDoncila:
@msschwartz21 I've merged your changes and started added some division tests. Right now I'm just asserting what's currently happening, but I actually think we should update our division correcting logic to capture the frame-skip edges. In the case of `div_parent_gap`, this should be easy enough just by changing `_get_succ_by_t` to compute the target time and then iterate through successors until it hits that time. 

For the `div_daughter_gap` graph, it's a little trickier because we would need to start checking `WC_DIV`s. However, I think it could be limited to only checking `WC_DIV`s with outgoing frame-skip edges, since that's the only way a parent could be in the same frame and be correct after a shift.

With track overlap too it's not really clear that we should penalize some of these cases e.g. `gap_close_gt_gap` or `gap_close_pred_gap`. However, the code currently checks specifically for the same set of edges to be present in both tracks (it just updates the "overlapping" edges to their matched node counterparts). 

I'm not sure how we want to proceed here:

- We could add the tests for what's currently happening, document the behaviour, and make an issue to improve our handling of skip-edges OR
- We could update all the code in this PR to exhibit the desired behaviour and then add appropriate tests

Regardless of which we choose, I do think we should update the code so that, broadly speaking, a skip-frame edge `u -> v` is identical to a set of edges `u' -> ... -> v'` if `u'` is a direct predecessor of `v'`. I'm tempted to add tests for current behaviour and/or just document it, and then follow up with smaller PRs that fixes the issues. 

Thoughts?

msschwartz21:
> I'm tempted to add tests for current behaviour and/or just document it, and then follow up with smaller PRs that fixes the issues.

I agree. Let's finish up testing the current behavior and get issues opened for the things we want to change. The updates you proposed to the division code seem reasonable and like they wouldn't be too big of a deal.

> Regardless of which we choose, I do think we should update the code so that, broadly speaking, a skip-frame edge u -> v is identical to a set of edges u' -> ... -> v' if u' is a direct predecessor of v'.

This seems reasonable, but I'll need to think about it about other cases where this might come up.

DragaDoncila:
@napari-bot update constraints

DragaDoncila:
ok @msschwartz21 I've updated all the tests and added some more documentation about the gap-closing edges. I made a dedicated page for Gap-Closing under Track Errors, mainly to aid discoverability in the case where a user is confused about the metrics they've computed being worse than they might expect, and also to give us a place to document any future behaviour changes. I also opened #211 to track our remaining work here.

I expanded in each Track Errors page on the effect of gap-closing edges, except for the ctc errors where I notice we don't discuss the induced graph or anything. I'm hoping that it's sufficient information for users to understand that a gap-closing edge will be treated the same as any other edge, but happy to expand as you see fit.

Lmk what you think!

msschwartz21:
@DragaDoncila This is great! Especially all the documentation 😂 

I opened another issue capturing the division related work which I think we should do #214 

My first impression is that this PR is ready to go, but I need to do a more careful pass on the tests which I'll do later today or tomorrow. Thank you for all your work on this one!

msschwartz21:
Before this gets merged, can we add a new group to the test coverage plot? The new functions just need to added to [this](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/9ea54547aaaaf71e2d4d9d797b53ad0a58bf0448/scripts/test_case_report.py#L33) dictionary with a new "gap closing" group.

msschwartz21 (approved):
I took a second pass through the tests and everything looks good! I opened #226 to take care of the change to the test case coverage plot, but once that's merged in this PR should be good to go!

DragaDoncila:
Merging this now as we merged in #226 and tests are passing!

PR #141, opened 2024-01-24
Check empty subgraph

cmalinmayor:
# Proposed Change
Fix bug related to empty induced subgraph when computing CTC errors. Also now throws an error in TrackingGraph.get_subgraph() when empty nodes are provided, which I'm not sure is the correct behavior, but an "empty" TrackingGraph is poorly defined (e.g. need start and end frame determined by data).

# Types of Changes
- [ ] Documentation update
- [x] Tests and benchmarks

- [ ] Loaders
- [ ] Matchers
- [x] Track Errors
- [ ] Metrics
- [x] Core functionality (e.g. `TrackingGraph`, `run_metrics`, `cli`, etc.)

# Checklist

### DIFF
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -203,10 +203,6 @@ class TrackingGraph:
                 if edge_flag in attrs and attrs[edge_flag]:
                     self.edges_by_flag[edge_flag].add(edge)
 
-        # Store first and last frames for reference
-        self.start_frame = min(self.nodes_by_frame.keys())
-        self.end_frame = max(self.nodes_by_frame.keys()) + 1
-
         # Record types of annotations that have been calculated
         self.division_annotations = False
         self.node_errors = False
@@ -347,9 +343,6 @@ class TrackingGraph:
             new_trackgraph.edges_by_flag[edge_flag] = self.edges_by_flag[
                 edge_flag
             ].intersection(new_trackgraph.edges)
-
-        new_trackgraph.start_frame = min(new_trackgraph.nodes_by_frame.keys())
-        new_trackgraph.end_frame = max(new_trackgraph.nodes_by_frame.keys()) + 1
 
         return new_trackgraph
 
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -205,8 +205,12 @@ class TrackingGraph:
                     self.edges_by_flag[edge_flag].add(edge)
 
         # Store first and last frames for reference
-        self.start_frame = min(self.nodes_by_frame.keys())
-        self.end_frame = max(self.nodes_by_frame.keys()) + 1
+        if len(self.nodes_by_frame) == 0:
+            self.start_frame = None
+            self.end_frame = None
+        else:
+            self.start_frame = min(self.nodes_by_frame.keys())
+            self.end_frame = max(self.nodes_by_frame.keys()) + 1
 
         # Record types of annotations that have been calculated
         self.division_annotations = False
@@ -328,8 +332,12 @@ class TrackingGraph:
                 edge_flag
             ].intersection(new_trackgraph.edges)
 
-        new_trackgraph.start_frame = min(new_trackgraph.nodes_by_frame.keys())
-        new_trackgraph.end_frame = max(new_trackgraph.nodes_by_frame.keys()) + 1
+        if len(new_trackgraph.nodes_by_frame) == 0:
+            new_trackgraph.start_frame = None
+            new_trackgraph.end_frame = None
+        else:
+            new_trackgraph.start_frame = min(new_trackgraph.nodes_by_frame.keys())
+            new_trackgraph.end_frame = max(new_trackgraph.nodes_by_frame.keys()) + 1
 
         return new_trackgraph
 
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -316,13 +316,19 @@ class TrackingGraph:
         return [self.get_subgraph(g) for g in nx.weakly_connected_components(graph)]
 
     def get_subgraph(self, nodes: Iterable[Hashable]) -> TrackingGraph:
-        """Returns a new TrackingGraph with the subgraph defined by the list of nodes
+        """Returns a new TrackingGraph with the subgraph defined by the list of nodes.
 
         Args:
-            nodes (list): List of node ids to use in constructing the subgraph
-        """
+            nodes (list): A non-empty list of node ids to use in constructing the subgraph
 
+        Raises:
+            ValueError if nodes is empty
+        """
         new_graph = self.graph.subgraph(nodes).copy()
+        logger.debug(f"Subgraph has {new_graph.number_of_nodes()} nodes")
+        if new_graph.number_of_nodes() == 0:
+            raise ValueError("Cannot pass an empty set of nodes to subgraph")
+
         new_trackgraph = copy.deepcopy(self)
         new_trackgraph.graph = new_graph
         for frame, nodes_in_frame in self.nodes_by_frame.items():
@@ -330,6 +336,7 @@ class TrackingGraph:
             if new_nodes_in_frame:
                 new_trackgraph.nodes_by_frame[frame] = new_nodes_in_frame
             else:
+                logger.debug("Frame {frame} has no nodes in subgraph")
                 del new_trackgraph.nodes_by_frame[frame]
 
         for node_flag in NodeFlag:
@@ -339,7 +346,7 @@ class TrackingGraph:
         for edge_flag in EdgeFlag:
             new_trackgraph.edges_by_flag[edge_flag] = self.edges_by_flag[
                 edge_flag
-            ].intersection(nodes)
+            ].intersection(new_trackgraph.edges)
 
         new_trackgraph.start_frame = min(new_trackgraph.nodes_by_frame.keys())
         new_trackgraph.end_frame = max(new_trackgraph.nodes_by_frame.keys()) + 1
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -315,15 +315,10 @@ class TrackingGraph:
         """Returns a new TrackingGraph with the subgraph defined by the list of nodes.
 
         Args:
-            nodes (list): A non-empty list of node ids to use in constructing the subgraph
-
-        Raises:
-            ValueError if nodes is empty
+            nodes (list): A list of node ids to use in constructing the subgraph
         """
         new_graph = self.graph.subgraph(nodes).copy()
         logger.debug(f"Subgraph has {new_graph.number_of_nodes()} nodes")
-        if new_graph.number_of_nodes() == 0:
-            raise ValueError("Cannot pass an empty set of nodes to subgraph")
 
         new_trackgraph = copy.deepcopy(self)
         new_trackgraph.graph = new_graph
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -311,7 +311,6 @@ class TrackingGraph:
             nodes (list): A list of node ids to use in constructing the subgraph
         """
         new_graph = self.graph.subgraph(nodes).copy()
-        logger.debug(f"Subgraph has {new_graph.number_of_nodes()} nodes")
 
         new_trackgraph = copy.deepcopy(self)
         new_trackgraph.graph = new_graph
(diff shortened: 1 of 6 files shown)


src/traccuracy/track_errors/_ctc.py
```diff
@@ -87,9 +88,11 @@ def get_edge_errors(matched_data: Matched):
...
         get_vertex_errors(matched_data)
 
-    induced_graph = comp_graph.get_subgraph(
-        comp_graph.get_nodes_with_flag(NodeFlag.TRUE_POS)
-    ).graph
+    comp_tp_nodes = comp_graph.get_nodes_with_flag(NodeFlag.TRUE_POS)
+    if len(comp_tp_nodes) == 0:
+        induced_graph = nx.DiGraph()
+    else:
+        induced_graph = comp_graph.get_subgraph(comp_tp_nodes).graph
```
DragaDoncila, line 95: Since we're accessing the networkx graph anyway using `.graph`, what do you think about just keeping 

`induced_graph = comp_graph.graph.subgraph(comp_tp_nodes)`

here? This would give us an empty graph for the 0 case without having to do the if statement, and would also avoid the extra overhead of getting the `TrackingGraph` subgraph just to access the nx attribute anyway

tests/track_errors/test_ctc_errors.py
```diff
@@ -134,3 +134,53 @@ def test_assign_edge_errors_semantics():
...
+
+    mapping = [
+        (1, 1),
+        (5, 1),
+        (2, 2),
+        (5, 2),
+    ]
+
+    matched_data = Matched(TrackingGraph(gt), TrackingGraph(comp), mapping)
+    get_edge_errors(matched_data)
```
DragaDoncila, line 186: would be good here to actually assert on the counts of each error

DragaDoncila (commented):
Thanks for fixing this one @cmalinmayor - left just a couple of comments

msschwartz21 (commented):
Generally this seems reasonable to me and I'm in agreement with Drag's comments. My only question is if we really want to fail and throw a ValueError in the case where the subgraph is empty. I'm mostly thinking about downstream uses where maybe it would be easier to just return None and then it's the responsibility of the downstream code to handle that case if it appears.

DragaDoncila:
> My only question is if we really want to fail and throw a ValueError in the case where the subgraph is empty.

Yeah I had this thought too actually. I wonder if we should just mimic the networkx behaviour again here, so returning an empty graph. I think that's also the correct subset result for an intersection with an empty set. Based on both of these things it's probably also the "expected" result.

src/traccuracy/_tracking_graph.py
```diff
@@ -316,20 +316,27 @@ def get_connected_components(self) -> list[TrackingGraph]:
...
-            nodes (list): List of node ids to use in constructing the subgraph
-        """
+            nodes (list): A non-empty list of node ids to use in constructing the subgraph
 
+        Raises:
+            ValueError if nodes is empty
+        """
         new_graph = self.graph.subgraph(nodes).copy()
+        logger.debug(f"Subgraph has {new_graph.number_of_nodes()} nodes")
+        if new_graph.number_of_nodes() == 0:
```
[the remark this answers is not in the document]
cmalinmayor: The only issue is you can techncially pass nodes that aren't in the graph to subgraph, and it will take the intersection of the passed list of nodes and those in the graph. So a len(nodes) != 0 does not necessarily mean the subgraph will have nodes. This may not matter if we allow empty TrackingGraphs as discussed below

cmalinmayor:
@msschwartz21 @DragaDoncila Thanks for the feedback. The main issue with returning an empty TrackingGraph was the start_frame and end_frame attributes, which assume the graph is not empty in our current implementation. After some attempts to fix this by setting them to None if its empty and running into typing issues, I decided to try removing these attributes. Let me know what you think. It means we can create an empty TrackingGraph easily, can avoid checking if the subgraph is empty and just return an empty TrackingGraph, and can get the start_frame and end_frame by sorting `TrackingGraph.nodes_by_frame.keys()` when we need them.

src/traccuracy/_tracking_graph.py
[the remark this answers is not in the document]
DragaDoncila: Ah I see - ok let's wait on that discussion and then resolve this.

DragaDoncila:
> The main issue with returning an empty TrackingGraph was the start_frame and end_frame attributes, which assume the graph is not empty in our current implementation. After some attempts to fix this by setting them to None if it's empty and running into typing issues, I decided to try removing these attributes.

Hmm @cmalinmayor I wonder what the danger is of just keeping the `min_frame` and `max_frame` of the original `TrackingGraph`? Does the `segmentation` (if it's there) get given to the subgraph? If so then you could think of this as the same "problem" instance, but just with a different set of detections now (i.e. none) - could be useful?  We already allow frames with no detections, so this would just be the extreme case.

cmalinmayor:
@DragaDoncila I added asserts on the expected behavior in my minimal test case. Turns out, I don't understand the expected behavior 🤷‍♀️ There are two inconsistencies between what I would expect and what happens in the example. 

1. What should happen to the ground truth nodes if the comp nodes are marked Non-split? I know what I did [here](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/bd12224e49c4c8370e1ecbb15b97d5fc448754c0/tests/track_errors/test_ctc_errors.py#L199) was wrong, because the non-split attribute can only be on the computed graph. But when I print all the attributes on the gt nodes for the example:

```
gt         1 - 2 - 3
           4 - 5 - 6
comp       1 - 2
matching [ (1, 1), (4, 1), (2, 2), (5, 2) ]

GT NODES
1 {'t': 0, 'x': 1, 'y': 1, <NodeFlag.FALSE_NEG: 'is_fn'>: False}
2 {'t': 1, 'x': 1, 'y': 1, <NodeFlag.FALSE_NEG: 'is_fn'>: False}
4 {'t': 0, 'x': 0, 'y': 1, <NodeFlag.FALSE_NEG: 'is_fn'>: True} *(huh?)
5 {'t': 1, 'x': 0, 'y': 1, <NodeFlag.FALSE_NEG: 'is_fn'>: False}
3 {'t': 2, 'x': 1, 'y': 1, <NodeFlag.FALSE_NEG: 'is_fn'>: True} *(expected)
6 {'t': 2, 'x': 0, 'y': 1, <NodeFlag.FALSE_NEG: 'is_fn'>: True} *(expected)
```
One of the four GT nodes matched to a non-split node is a false negative, and the others are not. Which is correct?

2. If the ground truth edges between nodes matched to "non-split" nodes are false negative edges as previously discussed, why are the comp edges not false positives? [This](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/bd12224e49c4c8370e1ecbb15b97d5fc448754c0/tests/track_errors/test_ctc_errors.py#L194) assert fails. All edge attributes printed for the example:
```
gt         1 - 2 - 3
           4 - 5 - 6
comp       1 - 2
matching [ (1, 1), (4, 1), (2, 2), (5, 2) ]

COMP EDGE
(1, 2) {<EdgeFlag.FALSE_POS: 'is_fp'>: False, <EdgeFlag.WRONG_SEMANTIC: 'is_wrong_semantic'>: False, <EdgeFlag.INTERTRACK_EDGE: 'is_intertrack_edge'>: False}
GT EDGES
(1, 2) {<EdgeFlag.FALSE_NEG: 'is_fn'>: True, <EdgeFlag.INTERTRACK_EDGE: 'is_intertrack_edge'>: False}
(2, 3) {<EdgeFlag.FALSE_NEG: 'is_fn'>: True, <EdgeFlag.INTERTRACK_EDGE: 'is_intertrack_edge'>: False}
(4, 5) {<EdgeFlag.FALSE_NEG: 'is_fn'>: True, <EdgeFlag.INTERTRACK_EDGE: 'is_intertrack_edge'>: False}
(5, 6) {<EdgeFlag.FALSE_NEG: 'is_fn'>: True, <EdgeFlag.INTERTRACK_EDGE: 'is_intertrack_edge'>: False}
```

cmalinmayor:
@DragaDoncila @msschwartz21 how do you feel about merging these fixes ASAP even though we have uncovered other questions by adding more thorough tests? I really need the key error and the empty subgraph issues fixed. At Morgan's suggestion, I added back in the start_frame and end_frame and set them to None if the TrackingGraph is empty (both at initialization and subgraph steps).

DragaDoncila:
@cmalinmayor I'm happy to merge the fix even though we've uncovered more questions! Are the currently failing tests a result of the new assertions you tried to make? If so, it would be good to either remove them to get CI green or maybe even better, mark them as `skip` so that we have them but aren't running them and we can just unmark them in the PR that addresses the new inconsistencies you've found.

> What should happen to the ground truth nodes if the comp nodes are marked Non-split? 

According to the paper, the induced subgraph that we compute after we compute the node matching is formed strictly of the "uniquely matching computed vertices and their incident edges", so all GT nodes associated with an NS node should be deleted in the subgraph. This kinda implies that they should **both** be marked FN...?

> If the ground truth edges between nodes matched to "non-split" nodes are false negative edges as previously discussed, why are the comp edges not false positives?

This one is super weird but our induced subgraph in this case would be empty (we have two non-split nodes so they both get removed and by extension their edge also gets removed). The paper says "We define the set of redundant edges (FP edges) in the computed graph as the induced subgraph edges without any counterpart in the ground truth graph", so, because that single edge doesn't exist in the induced subgraph, it doesn't get marked as FP and you just get penalized once for your FN edges from the GT...

cmalinmayor:
> This one is super weird but our induced subgraph in this case would be empty (we have two non-split nodes so they both get removed and by extension their edge also gets removed). The paper says "We define the set of redundant edges (FP edges) in the computed graph as the induced subgraph edges without any counterpart in the ground truth graph", so, because that single edge doesn't exist in the induced subgraph, it doesn't get marked as FP and you just get penalized once for your FN edges from the GT...

Wouldn't all edges adjacent to NS nodes get removed, even if one endpoint is still there? Then we never get penalized for edges next to NS nodes being FP? If that's expected, then I can change the assert for the edges and just kick the can on the node situation.

DragaDoncila:
> Wouldn't all edges adjacent to NS nodes get removed, even if one endpoint is still there? 

Yeah that's correct.

> Then we never get penalized for edges next to NS nodes being FP?

I think that's also correct. If you have an NS node, it gets removed from the induced graph, as do all of its adjacent edges. So I guess for NS nodes you **only** get penalized for their NS weight plus any FN edges that are now created by the fact that you've removed edges adjacent to the NS node.

> If that's expected, then I can change the assert for the edges and just kick the can on the node situation.

I think that sounds ok

src/traccuracy/_tracking_graph.py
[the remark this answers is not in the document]
cmalinmayor: This check has been removed, since we now allow empty TrackingGraphs

tests/track_errors/test_ctc_errors.py
cmalinmayor: Let's count this as semi-resolved 😅

src/traccuracy/matchers/_ctc.py
```diff
@@ -51,8 +51,11 @@ def _compute_mapping(self, gt_graph: TrackingGraph, pred_graph: TrackingGraph):
         if mask_gt.shape != mask_pred.shape:
             raise ValueError("Segmentation shapes must match between gt and pred")
 
-        mapping = []
+        mapping: list[tuple] = []
         # Get overlaps for each frame
+        if gt.start_frame is None or gt.end_frame is None:
```
msschwartz21, line 56: Just checking my understanding, could this also be an `and` because we wouldn't expect the start or end frame to be None without the other also being none?

msschwartz21 (approved):
Left one question just checking my understanding, but otherwise this looks good to go!

src/traccuracy/matchers/_ctc.py
cmalinmayor: Right, because they are always set together

tests/track_errors/test_ctc_errors.py
```diff
@@ -134,3 +134,69 @@ def test_assign_edge_errors_semantics():
...
+    matched_data = Matched(TrackingGraph(gt), TrackingGraph(comp), mapping)
+    get_vertex_errors(matched_data)
+    get_edge_errors(matched_data)
+
+    for node in comp.nodes:
+        assert comp.nodes[node][NodeFlag.NON_SPLIT]
+    for edge in comp_edges:
+        assert not comp.edges[edge][EdgeFlag.FALSE_POS]
+
+    if False:  # TODO: Fix this in a separate PR
```
DragaDoncila, line 194: @cmalinmayor can you link back to the original comment here, just in case we forget what we had to fix

https://github.com/Janelia-Trackathon-2023/traccuracy/pull/141#issuecomment-2265990197

DragaDoncila (approved):
I haven't pulled this down to play with it as I'm away until next Thursday, but I had a look through and played before so looks good to me, just left a small comment!

PR #149, opened 2024-08-02
Check for missing computed endpoints in induced graph

### LINKED ISSUE #137: Key Error in CTC edge error computation with non-split vertices
* traccuracy version: 0.1.0
* Python version: 3.11.6
* Operating System: macOS

# Description

I was running the CTC metrics on an example and kept getting exceptions. First, when the induced graph was empty (no TP nodes), the CTC metrics raised a confusing error. I fixed this in b365b8f330d077b3f465a0004882787893eda532, but still am getting a KeyError when looking up the expected computed edge endpoints in the gt_comp_mapping. I have created a minimal test case in 0802b84fa9fc9dce58a45ca37b44e27531890a99 that I believe replicates the issue.

My understanding of the issue is as follows: We are looking up the expected computed edge in the gt_comp_mapping, which was constructed to only include mappings where the computed node was included in the "induced graph". The “induced graph” only includes TP computed nodes, and not ones labeld Non-split. The if statement here 
https://github.com/Janelia-Trackathon-2023/traccuracy/blob/0802b84fa9fc9dce58a45ca37b44e27531890a99/src/traccuracy/track_errors/_ctc.py#L140-L145
only considers if the ground truth node is a false negative. If the ground truth node was matched to a non-split node, it is not labeled false negative, so we get past this if statement and try and identify the expected computed edge. However, the matched non-split computed node will not be in the induced graph and then not in the mapping, hence the key error. 

@DragaDoncila I need your guidance on how to fix this. Is a GT edge matched to a predicted edge where one end is a non-split vertex considered a False negative? Does the above explanation even make sense or is something else fishy going on here?

https://github.com/Janelia-Trackathon-2023/traccuracy/blob/0802b84fa9fc9dce58a45ca37b44e27531890a99/tests/track_errors/test_ctc_errors.py#L139-L186

# Severity
- [x] Unusable
- [ ] Very minor

cmalinmayor:
# Proposed Change
Closes #137 by implementing the change that @DragaDoncila suggested

# Types of Changes
- [ ] Documentation update
- [ ] Tests and benchmarks

- [ ] Loaders
- [ ] Matchers
- [x] Track Errors
- [ ] Metrics

# Checklist

### DIFF
--- a/src/traccuracy/track_errors/_ctc.py
+++ b/src/traccuracy/track_errors/_ctc.py
@@ -141,8 +141,12 @@ def get_edge_errors(matched_data: Matched):
             gt_graph.set_flag_on_edge(edge, EdgeFlag.FALSE_NEG, True)
             continue
 
-        source_comp_id = gt_comp_mapping[source]
-        target_comp_id = gt_comp_mapping[target]
+        source_comp_id = gt_comp_mapping.get(source, None)
+        target_comp_id = gt_comp_mapping.get(target, None)
+
+        if source_comp_id is None or target_comp_id is None:
+            gt_graph.set_flag_on_edge(edge, EdgeFlag.FALSE_NEG, True)
+            continue
 
         expected_comp_edge = (source_comp_id, target_comp_id)
         if expected_comp_edge not in induced_graph.edges:


cmalinmayor:
This is actually covered in #141

Issue #100, opened 2023-11-02
[BUG] Frame buffer specification always includes divisions from delta 0 onwards


KristinaUlicna:
Playing with the `ctc.ipynb` data, I'm trying to change the `metrics_kwargs` argument to include a different range of frame buffers to compute the metrics for, as follows:

```
iou_results = run_metrics(
    gt_data=gt_data, 
    pred_data=pred_data, 
    matcher=IOUMatched, 
    matcher_kwargs=dict(
        iou_threshold=iou
    ),
    metrics=[DivisionMetrics], 
    metrics_kwargs=dict(
        frame_buffer=(1, )
    ),
)
pp.pprint(iou_results["DivisionMetrics"])
```

where `iou` is somewhere between 0.3-0.9. 

The output always includes the results of `frame_buffer = 0`, even when not specified:

```
IoU Threshold : 0.9
Matching frames: 100%|██████████| 92/92 [00:05<00:00, 16.02it/s]
0.9
{   'Frame Buffer 0': {   'Division F1': 0.4318181818181818,
                          'Division Precision': 0.4175824175824176,
                          'Division Recall': 0.4470588235294118,
                          'False Negative Divisions': 94,
                          'False Positive Divisions': 106,
                          'Mitotic Branching Correctness': 0.2753623188405797,
                          'True Positive Divisions': 76},
    'Frame Buffer 1': {   'Division F1': 0.4318181818181818,
                          'Division Precision': 0.4175824175824176,
                          'Division Recall': 0.4470588235294118,
                          'False Negative Divisions': 94,
                          'False Positive Divisions': 106,
                          'Mitotic Branching Correctness': 0.2753623188405797,
                          'True Positive Divisions': 76}}
```

I'm not sure if this is a **bug** _per se_, but it may be beneficial to change the argument type from `tuple[int]` to an `max_int` in this case, where 2 would compute all frame buffers up to 2, _i.e._ 0, 1 as well as 2.

bentaculum:
Hi @msschwartz21, have you already had a chance to look this one? I just ran into this as well. As for Kristina above, for me different frame buffers, e.g. 0 vs 100, are leading to the same metrics, seems to be a bug.

msschwartz21:
Hey Ben, I thought the issue here was that a frame buffer of 0 is always calculated even when another value is requested. Is that the issue you are seeing or is it something else?

bentaculum:
[...]

PR #155, opened 2024-08-06
Fix frame buffer bug where predecessors were not checked properly

cmalinmayor:
# Proposed Change
@bentaculum I also saw issues with the frame buffer not updating when I expected it to. Upon closer inspection, there was a bug in how we were checking the successors. This PR fixes that bug. Sorry!

Previous discussion of possible frame buffer issue https://github.com/Janelia-Trackathon-2023/traccuracy/issues/100#issuecomment-1852330818

# Types of Changes
- [ ] Documentation update
- [ ] Tests and benchmarks

- [ ] Loaders
- [ ] Matchers
- [x] Track Errors
- [ ] Metrics

# Checklist

### DIFF
--- a/src/traccuracy/track_errors/divisions.py
+++ b/src/traccuracy/track_errors/divisions.py
@@ -174,6 +174,9 @@ def _correct_shifted_divisions(matched_data: Matched, n_frames=1):
     fp_divs = g_pred.get_nodes_with_flag(NodeFlag.FP_DIV)
     fn_divs = g_gt.get_nodes_with_flag(NodeFlag.FN_DIV)
 
+    gt_to_pred_dict = dict(mapper)
+    pred_to_gt_dict = {pred: gt for (gt, pred) in mapper}
+
     # Compare all pairs of fp and fn
     for fp_node, fn_node in itertools.product(fp_divs, fn_divs):
         correct = False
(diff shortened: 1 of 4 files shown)


msschwartz21:
I'm going to go in and fix the test case for this so that the tests would have actually caught this bug...

bentaculum:
Nice catch! Can we generally describe the influence of this fix on the division metrics calculated up to this point? Are old metrics too optimistic/pessimistic?

cmalinmayor:
@bentaculum The old version incorrectly penalized some (most?) correct divisions when frame_buffer > 0. Specifically, it checked if the successors had the same ID, not if the successors were matched, so if your prediction and candidate nodes had different IDs, it would never count an offset division as correct. For linking only experiments with GT detections, it probably won't change anything, but for experiments with predicted detections, performance should improve for division metrics when frame buffer > 0.

DragaDoncila:
@cmalinmayor took a look at this PR and if I'm understanding correctly, the bug was that when checking successors, we were directly using the graph node IDs rather than converting them via the mapper. And the test was passing because our `division_graphs` are using identical node IDs right? So if we just tweak the `get_division_graphs` util to use different node IDs that should be a sufficient test that fails on main and passes with this PR?

msschwartz21:
@DragaDoncila that's correct! I'll fix the test case in the next week or so. It's an easy fix but life has been busy 😂

DragaDoncila:
@msschwartz21 no rush at all, just making sure I understand the change!

DragaDoncila (approved):
Looks good to me! Thanks for editing the tests @msschwartz21

PR #160, opened 2024-09-10
Group PRs in release notes based on labels

msschwartz21:
Prettier release notes :) Not sure how to test it until we run a release

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,39 @@
+changelog:
+  categories:
+    - title: '✨ Features'
+      labels:
+        - 'feature'
+        - 'enhancement'
+    - title: '🚀 Performance'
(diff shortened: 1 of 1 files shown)


DragaDoncila:
@msschwartz21 I think we could test by tagging a release candidate, so you could tag `v0.1.2rc0` instead of just `v0.1.2`. This will go through the full release workflow (including generating release notes), but users installing traccuracy would still get `v0.1.1` unless they run `pip install --pre traccuracy`. If we're not happy with the format we can keep bumping the rc, and then when we are happy, we just tag `v0.1.2` and run the full release.

DragaDoncila (approved):
Approving optimistically and we can edit once we release an rc!

PR #170, opened 2024-11-25
Metrics validate that the matcher is supported

### LINKED ISSUE #83: Feature: Validation of Matchers and Metrics that work together

msschwartz21:
Closes #83 

Ultimately we are embracing a very manual/hard-coded approach. Each metric has to implement its own `_validate_matcher` function that can check both the name of the matcher and any associated parameters that may be relevant for that metric. For example, `IOUMatcher` is only valid for division metrics if `one_to_one=True`. 

# Types of Changes
- New feature or enhancement

- Matchers
- Metrics

# Checklist

### DIFF
--- a/src/traccuracy/metrics/_base.py
+++ b/src/traccuracy/metrics/_base.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from importlib.metadata import version
 from typing import TYPE_CHECKING
+import warnings
 
 if TYPE_CHECKING:
     from traccuracy.matchers._base import Matched
@@ -15,11 +16,12 @@ class Metric(ABC):
     Kwargs should be specified in the constructor
     """
 
-    # Mapping criteria
-    needs_one_to_one = False
-    supports_one_to_many = False
-    supports_many_to_one = False
-    supports_many_to_many = False
+    @abstractmethod
+    def _validate_matcher(self, matched: Matched) -> bool:
+        """Verifies that the matched meets the assumptions of the metric
+        Returns True if matcher is valid and False if matcher is not valid"""
+
+        raise NotImplementedError
 
     @abstractmethod
     def _compute(self, matched: Matched) -> dict:
@@ -36,7 +38,7 @@ class Metric(ABC):
         """
         raise NotImplementedError
 
-    def compute(self, matched: Matched) -> Results:
+    def compute(self, matched: Matched, override_matcher: bool = False) -> Results:
         """The compute methods of Metric objects returns a Results object populated with results
         and associated metadata
 
@@ -46,6 +48,15 @@ class Metric(ABC):
         Returns:
             Results: Object containing metric results and associated pipeline metadata
         """
+        if override_matcher:
+            warnings.warn("Overriding matcher/metric validation may result in "
+                         "unpredictable/incorrect metric results")
+        
+        valid_matcher = self._validate_matcher(matched)
+        if not valid_matcher:
+            raise TypeError(
+                "The matched data uses a matcher that does not meet the requirements "
+                "of the metric. Check the documentation for the metric for more information.")
 
         res_dict = self._compute(matched)
 
--- a/src/traccuracy/metrics/_ctc.py
+++ b/src/traccuracy/metrics/_ctc.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from traccuracy._tracking_graph import EdgeFlag, NodeFlag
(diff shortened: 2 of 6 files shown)


src/traccuracy/metrics/_base.py
```diff
@@ -46,6 +48,19 @@ def compute(self, matched: Matched) -> Results:
         Returns:
             Results: Object containing metric results and associated pipeline metadata
         """
+        if override_matcher:
```
cmalinmayor, line 51: Is the intent of this to allow skipping the validation? I think it just prints the warning and doesn't actually skip it

src/traccuracy/metrics/_ctc.py
```diff
@@ -34,6 +34,15 @@ def __init__(
             "ws": edge_ws_weight,
         }
 
+    def _validate_matcher(self, matched: Matched) -> bool:
+        valid_matchers = {"IOUMatcher", "CTCMatcher"}
+        name = matched.matcher_info["name"]
+
+        if name in valid_matchers:
```
cmalinmayor, line 41: Could just `return name in valid_matchers` but this is a style thing

src/traccuracy/metrics/_divisions.py
```diff
@@ -63,11 +64,20 @@ class DivisionMetrics(Metric):
...
-
     def __init__(self, max_frame_buffer=0):
         self.frame_buffer = max_frame_buffer
 
+    def _validate_matcher(self, matched: Matched) -> bool:
+        "Matcher must be one to one"
+        name = matched.matcher_info["name"]
+
+        if name == "IOUMatcher":
+            flag = matched.matcher_info["one_to_one"]
```
cmalinmayor, line 75: Can you also check if IOU threshold is > .5? Because then the IOU matcher is implicitly one-to-one (and IIRC it is is more efficient to not actually set the flag, since I think adds an extra step of computation after the initial matching)

src/traccuracy/metrics/_track_overlap.py
```diff
@@ -57,11 +59,18 @@ class TrackOverlapMetrics(Metric):
 
     """
 
-    supports_many_to_one = True
-
     def __init__(self, include_division_edges: bool = True):
         self.include_division_edges = include_division_edges
 
+    def _validate_matcher(self, matched: Matched) -> bool:
```
cmalinmayor, line 65: I went down an issue rabbit hole and I believe that it is correct that track overlap works with all matchers. But, perhaps we should double/triple check? Or, just reassure me that you already did so 😄

cmalinmayor (changes requested):
Generally looks good! Thanks for implementing :) just left a few comments inline

src/traccuracy/metrics/_base.py
msschwartz21: You are entirely correct and code review is a wonderful thing :joy:

src/traccuracy/metrics/_track_overlap.py
msschwartz21: My due diligence was limited, but I think based on what I've ready that all of our current matchers should be ok. I'm going to add to the docstring so that we have a note that the intention si to be many to one.

src/traccuracy/metrics/_divisions.py
```diff
@@ -70,13 +70,16 @@ def __init__(self, max_frame_buffer=0):
...
+        valid = False
 
         if name == "IOUMatcher":
-            flag = matched.matcher_info["one_to_one"]
-            if flag:
-                return True
+            if matched.matcher_info["one_to_one"]:
+                valid = True
+            # Threshold of greater than 0.5 ensures one to one
+            if matched.matcher_info["iou_threshold"] >= 0.5:
```
cmalinmayor, line 79: Super pedantic but probably should be strictly greater 😬

cmalinmayor (approved):
Approving with the caveat that it should be strictly greater than 0.5 for true one-to-one-ness :) but I don't feel the need to review again once that's fixed!

PR #172, opened 2024-12-10
Add tests cases to docs using example notebook

msschwartz21:
# Types of Changes
What types of changes does your code introduce? Delete those that do not apply.
- Documentation update

# Checklist
Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.

- [x] I have read the developer/contributing docs.
- [x] I have added tests that prove that my feature works in various situations or tests the bugfix (if appropriate).
- [x] I have checked that I maintained or improved code coverage.
- [x] I have checked the benchmarking action to verify that my changes did not adversely affect performance.
- [x] I have written docstrings and checked that they render correctly in the Read The Docs build (created after the PR is opened).
- [x] I have updated the general documentation including Metric descriptions and example notebooks if necessary.

PR #173, opened 2024-12-10
Test the IOU matcher using standardized segmentation test cases

msschwartz21:
~~Branch is based on #172~~

I left the tests for `match_iou` and `IOUMatched` as is because the tests are more about checking the input validation than actually testing all the iou matching edge cases. All of the edge cases should be captured in `Test__match_nodes`.

# Types of Changes
- Tests and benchmarks

- Matchers

# Checklist


msschwartz21:
This reminds me that we should probably figure out why our code coverage report is failing before we make a bunch of test changes... I'll look into it tomorrow.

tests/examples/segs.py
```diff
@@ -263,6 +294,37 @@ def undersegmentation_3d() -> tuple[np.ndarray, np.ndarray]:
     return gt, pred
 
 
+def no_overlap_3d() -> tuple[np.ndarray, np.ndarray]:
```
DragaDoncila, line 297: Maybe we want to declare test fixtures for these canonical examples in conftest.py? Makes them easier to use just by declaring them as arguments, and saves us an import.

tests/matchers/test_iou.py
```diff
@@ -15,146 +15,146 @@
...
+    )
+    def test_no_overlap(self, data_fxn):
+        gt, pred = data_fxn()
+        expected_matches = []
+        gtcells, rescells = _match_nodes(gt, pred)
+        assert Counter(expected_matches) == Counter(list(zip(gtcells, rescells)))
+
+    def test_input_error(self):
+        im = np.zeros((10, 10))
+        with pytest.raises(ValueError):
```
DragaDoncila, line 137: Maybe we can use the `match` argument for `raises` here to ensure the error message corresponds to the specific error we're expecting i.e. 

```suggestion
        with pytest.raises(ValueError, match="Threshold of 0 is not valid unless one_to_one is True"):
```

tests/matchers/test_iou.py
```diff
@@ -15,146 +15,146 @@
...
-    # even with threshold 0
-    gtcells, rescells = _match_nodes(im1, im2, threshold=0.0, one_to_one=True)
-    # Create match tuples
-    matches = list(zip(gtcells, rescells))
-    # Check that nothing is matched
-    assert len(matches) == 0
+    @pytest.mark.parametrize(
+        "data_fxn", [ex_segs.no_overlap_2d, ex_segs.no_overlap_3d], ids=["2D", "3D"]
+    )
+    def test_non_sequential(self, data_fxn):
```
DragaDoncila, line 144: Just for my understanding, I'm not sure what this is testing? Why does it matter if the segmentation IDs are high non-sequential numbers?

DragaDoncila (commented):
Thanks for simplifying these @msschwartz21 just left a couple of comments

tests/examples/segs.py
cmalinmayor: We had them as fixtures and then ran into issues with calling the functions outside the tests to make visualizations put in the docs, and we didn't want to duplicate the code. Plus for the graph examples we want to parameterize over the functions, and having some be fixtures and some not seemed inconsistent.

tests/matchers/test_iou.py
cmalinmayor: There were more comments on the old version - I think we had issues at some point where dummy nodes from the match matrix were appearing in the matched output even though they weren't in the graph, since we added rows in the matrix up to the highest node id and used the index to represent the id. Not sure if we still do that or not, since I know we made the implementation more efficient.

msschwartz21: Caroline is correct about the issue we were trying to catch. I can add back in some comments so it's clearer. I was inclined to leave it in as a test case since it was an issue in the past even if we may have eliminated the opportunity for the problem.

tests/examples/segs.py
DragaDoncila: Ahh got it, thanks for the additional context

tests/matchers/test_iou.py
DragaDoncila: Ok great, thanks for the extra context. Agreed a short docstring would be good, can even link to this comment chain for posterity

cmalinmayor (changes requested):
Looks good other than my organizational suggestion! I did run codecov to ensure all the standard segmentation examples were at least called for each matcher, and it came up green. It did show that we have the "nodes_from_segmentation" function in the examples file that is unused. I think the idea was to make full movies by chaining segmentations and then extracting the graph, but I'm fine to take it out or leave it in.

cmalinmayor (approved):
I think it looks great, thanks! More end to end tests would be nice but I think that is a new issue for later, I'm happy to merge this now.

PR #176, opened 2024-12-18
Add tests for CTC node and edge track errors

msschwartz21:
# Proposed Change
Test the CTC track errors against the standard tests cases

Currently a work in progress because some of the tests are failing either due to a real bug or due to a misunderstanding of the correct CTC behavior. See next comment for elaboration on outstanding questions.

# Types of Changes
- Tests and benchmarks

- Track Errors

# Checklist


msschwartz21:
@DragaDoncila while trying to write the tests using the standard test case I identified a couple issues that I need you to weigh in on. If you can tell me what the "correct" ctc behavior is, I'll fix either the ctc code or the tests accordingly.

# No prediction
image
If there is no GT, I think all these predicted edges should be marked as FP. Currently they are not annotated with any edge flag. They are not included in the induced graph (because there are no TP edges) and therefore not considered for any edge annotation. I think this is a straightforward change, but I want your sign off because it is going to change AOGM values.

# Edges with non split vertices
image
When there is an edge with with a non-split vertex, how should the edge be classified?
See [example 1](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/ed4e03a7efdbe6845e8e184faf0c1ee9f1da727e/tests/track_errors/test_ctc_errors.py#L477) and [example 2](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/ed4e03a7efdbe6845e8e184faf0c1ee9f1da727e/tests/track_errors/test_ctc_errors.py#L502) for more specifics in the test cases.

DragaDoncila:
Hello @msschwartz21 thanks for the ping, I've had a look at these cases, both from a theoretical perspective and by generating examples and evaluating them with the CTC .exes.

# No prediction

> Currently they are not annotated with any edge flag. They are not included in the induced graph (because there are no TP edges) and therefore not considered for any edge annotation.

This is correct as per the CTC metric. A track consisting entirely of FP vertices is penalized only for its vertices and not for its edges. In the example you've shown, the edges would not be annotated and the AOGM score would incur a penalty of 3 FP vertices.

# Edges with non split vertices

When there is an edge adjacent to an NS vertex that edge is deleted from the induced subgraph and will typically incur some form of FN edge penalty, but no additional FP edge penalty. For your examples above, left to right:

1. Incurs 1 NS vertex penalty and 1 FN edge penalty for edge (1, 2), because edge (4, 5) is deleted in the induced subgraph.
2. Incurs 1 NS vertex penalty and 2 FN edge penalties for edges (1, 7) and (2, 3) because both edges (4, 5) and (5, 6) get deleted in the induced subgraph.
3. Same as 1, incurs 1 NS vertex penalty and 1 FN edge penalty for edge (2, 3) because edge (5, 6) is deleted in the induced subgraph.

All of this means there **will** be edges present in the solution that are not present in the ground truth but still remain unannotated under the CTC metrics, because they've been counted as "vertex errors". This is "by design" as per the AOGM paper:

image

Let me know if that clears everything up or if there's still weirdness in the implementation! The whole thing is unintuitive but I think we're correct "as per the metric".

msschwartz21:
> I've had a look at these cases, both from a theoretical perspective and by generating examples and evaluating them with the CTC .exes.

@DragaDoncila you are my hero/MVP for doing this ^ 🙏 

I am doing my best to accept that this is the intended behavior for the metric and have updated the tests accordingly.

msschwartz21:
Note: I merged in #179 which made the diff a bit messy. Sorry about that. The bulk of the relevant changes for this PR were in the test files themselves.

tests/track_errors/test_ctc_errors.py
```diff
@@ -194,15 +272,257 @@ def test_ns_vertex_fn_edge():
...
+            assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is False
+
+    @pytest.mark.parametrize("t", [0, 1])
+    def test_fp_edge(self, t):
+        matched = self.prep_matched(ex_graphs.fp_edge_matched(t))
+
+        # All but one pred edges correct
+        for edge, attrs in matched.pred_graph.edges.items():
+            # false positive edge
+            if edge == (7, 8):
```
DragaDoncila, line 389: The if and else branches here are identical (because the FP edge is entirely made of FP nodes), so maybe we should remove this test and keep the one below that does have FP edges involved?

tests/track_errors/test_ctc_errors.py
```diff
@@ -194,15 +272,257 @@ def test_ns_vertex_fn_edge():
...
+        """
+        matched = self.prep_matched(ex_graphs.node_two_to_one(0))
+
+        # All pred edges correct
+        for attrs in matched.pred_graph.edges.values():
+            assert attrs.get(EdgeFlag.CTC_FALSE_POS) is False
+            assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is False
+            assert attrs.get(EdgeFlag.WRONG_SEMANTIC) is False
+
+        # First edge correct
```
DragaDoncila, line 435:
```suggestion
        # First edge false neg as it was removed from induced graph
```

tests/track_errors/test_ctc_errors.py
```diff
@@ -194,15 +272,257 @@ def test_ns_vertex_fn_edge():
...
+        assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is False
+
+        # Pred edges are intertrack and wrong semantic
+        attrs = matched.pred_graph.edges[(6, 8)]
+        assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is True
+        assert attrs.get(EdgeFlag.WRONG_SEMANTIC) is True
+
+        attrs = matched.pred_graph.edges[(6, 7)]
+        assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is True
+        assert attrs.get(EdgeFlag.WRONG_SEMANTIC) is False
```
DragaDoncila, line 494: maybe we can assert here that edge 6-7 is also FP

tests/track_errors/test_ctc_errors.py
```diff
@@ -194,15 +272,257 @@ def test_ns_vertex_fn_edge():
...
+        # Pred edge
+        attrs = matched.pred_graph.edges[(6, 8)]
+        assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is False
+        assert attrs.get(EdgeFlag.WRONG_SEMANTIC) is True
+
+        # Gt edges are just intertrack
+        attrs = matched.gt_graph.edges[(2, 4)]
+        assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is True
+        attrs = matched.gt_graph.edges[(2, 3)]
+        assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is True
```
DragaDoncila, line 509: Maybe we can also assert this edge is FN?

tests/track_errors/test_ctc_errors.py
```diff
@@ -194,15 +272,257 @@ def test_ns_vertex_fn_edge():
...
+
+    def test_wrong_child(self):
+        # Intertrack and wrong semantic check
+        matched = self.prep_matched(ex_graphs.wrong_child(1))
+
+        # One pred edge correct
+        attrs = matched.pred_graph.edges[(7, 8)]
+        assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is True
+        assert attrs.get(EdgeFlag.WRONG_SEMANTIC) is False
+        # Not wrong semantic b/c edge doesn't directly match to gt edge
```
DragaDoncila, line 519: So should this edge also be FP in that case? As both nodes are present in GT but the edge is not?

tests/track_errors/test_ctc_errors.py
```diff
@@ -194,15 +272,257 @@ def test_ns_vertex_fn_edge():
...
+        attrs = matched.pred_graph.edges[(7, 9)]
+        assert attrs.get(EdgeFlag.INTERTRACK_EDGE) is True
+        assert attrs.get(EdgeFlag.WRONG_SEMANTIC) is False
+
+        # Gt edges are just intertrack
```
DragaDoncila, line 524: Same here should edge 2-4 be FN?

DragaDoncila (commented):
Sorry for the delay in reviewing this one @msschwartz21! It's looking great, just left some comments - let me know if I misunderstood anything!

tests/test_tracking_graph.py
```diff
@@ -128,11 +128,11 @@ def test_constructor(nx_comp1):
     with pytest.raises(AssertionError):
```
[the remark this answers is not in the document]
msschwartz21: Might as well take care of it now

tests/track_errors/test_ctc_errors.py
msschwartz21: It's a little redundanct, but I'd actually like to keep this in here just for my own understanding/clarity. This case of fp edge is unintuitive to me so it's helpful to have an explicit assertion of the correct behavior. Caroline and I are also trying to have every error flagging method test against all available test cases, so this maintains that coverage.

I'll add to the comment so that it's clear why we are bothering.

DragaDoncila (approved):
Awesome, this looks excellent :heart:

PR #193, opened 2025-01-15
Incorrect daughter in division is labeled as wrong child

### LINKED ISSUE #190: Divisions with one incorrect daughter should be labeled as a FP (on pred) and FN (on gt)
While adding plots for our division documentation, I was looking more closely at the case when a predicted division has one correct child and one incorrect child. Currently we only classify the division node in the prediction as a FP, but I think the division node in the GT should also be classified as a false negative. 
Screenshot 2025-01-07 at 3 12 01 PM

msschwartz21:
# Proposed Change
Fixes #190 

# Types of Changes

- Track Errors

# Checklist


msschwartz21:
In one of our tests we load the predicted Hela tracks saved in `examples` and have hard coded division error value:
https://github.com/Janelia-Trackathon-2023/traccuracy/blob/87a5b03e7ffac833036c7840aef8172cbe896718/tests/metrics/test_divisions.py#L50-L52

This PR changes the number of False Positive divisions from 31 to 54. This is obviously specific to this particular solution graph, but it's still a larger change than I was expecting. @DragaDoncila @cmalinmayor Just want to double check that we are all on board with making this change? 

As an alternative we could introduce another error class for "Incorrect Division" which could be annotated in addition to or in place of the FP/FN flags in this specific case.

DragaDoncila:
@msschwartz21 I have no answers, but I have some questions :sweat_smile: 

I think based purely on the slightly vague descriptions in the literature, our interpretation is correct. A branching event with mismatched children should lead to an FP and FN division each being counted. I looked at these papers:

- Bise, R., Yin, Z., & Kanade, T. (2011, March). Reliable cell tracking by global data association. In 2011 IEEE international symposium on biomedical imaging: From nano to macro (pp. 1004-1010). IEEE. (https://www.ri.cmu.edu/pub_files/2011/0/ISBI2011_BISE_Final.pdf)
- Magnusson, K. E., Jaldén, J., Gilbert, P. M., & Blau, H. M. (2014). Global linking of cell tracks using the Viterbi algorithm. IEEE transactions on medical imaging, 34(4), 911-929. (https://pmc.ncbi.nlm.nih.gov/articles/PMC4765504/pdf/nihms753556.pdf)
- Ulman, V., Maška, M., Magnusson, K. E., Ronneberger, O., Haubold, C., Harder, N., ... & Ortiz-de-Solorzano, C. (2017). An objective comparison of cell-tracking algorithms. Nature methods, 14(12), 1141-1152. (https://www.nature.com/articles/nmeth.4473.pdf)

I wanted to check whether our numbers are the same as `py-ctcmetrics` results, at least as some form of a sense-check. I used our sample solution with the `CTCMatcher` (since I know this dataset will result in one-to-one matching) and computed `DivisionMetrics` on it, and then ran the `py-cycmetrics` evaluation as well. Our results don't match.

We report an F1 score of 0.697 while they report an F1 score (they term it as BC even though apparently BC according to the papers above should just be recall...?) of 0.76. Both recalls are the same, but the precisions differ. We report 0.613 while they report 0.717. The difference in precision is due strictly to the counts of FP divisions. We count 48 while they count only 30. I put it all in a table below. I'm not sure whether we're right, they're right, or we're both wrong, but I think we should investigate before we move forward with this.

| Event | Ours | Theirs |
|--------|--------|--------|
| TP Divs | 76 | 76 |
| FP Divs | **48** | **30** |
| FN Divs | 18 | 18 |
| Precision | **0.613** | **0.717** |
| Recall | 0.809 | 0.809 | 
| F1 | **0.697** | **0.76** |

cmalinmayor:
Thank you so much @DragaDoncila for doing the leg work!! Morgan and I (and Manan) had a discussion about this. We think the most flexible way to deal with this is to annotate a new class on both the GT and the pred graph (e.g. Wrong Child Division). Then the metrics can count them as FP or FN as they need. This does just kick the can down the road about how to compute the F1 score, and we will need to revisit this discussion then. But with this framework, we can honor the branching correctness definition by treating the WC-Divs only as FN (not FP), but still have the flexibility to make other more detailed metrics.
(Note: the WC-DIV class will/can also apply to divisions where both children are wrong but the division event is correct 😬).

DragaDoncila (approved):
I think just annotating the wrong children makes sense! It's also more useful information probably, than just a corresponding FN/FP division. Just left a comment for my own understanding

src/traccuracy/track_errors/divisions.py
```diff
@@ -72,12 +75,15 @@ def _find_pred_node_matches(pred_node):
...
+            cnt_pred = Counter(succ_pred)
+            if cnt_gt == cnt_pred:
                 g_gt.set_flag_on_node(gt_node, NodeFlag.TP_DIV, True)
                 g_pred.set_flag_on_node(pred_node, NodeFlag.TP_DIV, True)
-            # If daughters are at all mismatched, division is false negative
+            # If daughters are at all mismatched, division is a wrong child division
             else:
-                g_gt.set_flag_on_node(gt_node, NodeFlag.FN_DIV, True)
+                g_gt.set_flag_on_node(gt_node, NodeFlag.WC_DIV, True)
+                g_pred.set_flag_on_node(pred_node, NodeFlag.WC_DIV)
```
DragaDoncila, line 86: How come we sometimes need to provide the boolean value of the flag, like in the line above, but other times we just provide the flag itself with no value?

DragaDoncila:
Oh @msschwartz21 we probably also want to update the title of the PR

src/traccuracy/track_errors/divisions.py
msschwartz21: Valid question. The trues are left over from when we were being more explicit and densely annotating the graph. I'll clean it up so that's its consistent.

Issue #194, opened 2025-01-16
Matcher to Metric validation should occur after matching is complete


DragaDoncila:
In #170 we introduced a validator that checks whether a given matcher can be used to compute certain metrics, and immediately raises an error if this is not the case.

I think we should relax this check and perhaps only **warn** when the matcher/metrics combo is instantiated, and then raise only if, after matching, the condition is not met.

Right now, even though the vast majority of CTC GT matching will lead to one-to-one matching implicitly (because the GT annotations are often very small), users would not be able to `traccuracy` to compute e.g. division metrics.

msschwartz21:
I think the reason we didn't go down this path in the past was that we were trying to avoid actually checking the matching to see if it meets the conditions of the metric. However, you point out a pretty big limitation that may prevent people from using the non-CTC just because it will add an additional step. 

We could add a property to the `Matched` class that returns the type of matching, but only runs if the property is called. I'm assuming it will be somewhat expensive to do the check, but I could be missing a clever shortcut.

DragaDoncila:
> We could add a property to the Matched class that returns the type of matching, but only runs if the property is called. 

I think this is a good option! The other alternative is we could set the property to `True` by default, and then flip it in the matching step if we find an instance that invalidates it, which should mean we don't need an additional pass. I'll have to take a look at how doable that is but I think it should be possible when we construct the mapping itself.

> I'm assuming it will be somewhat expensive to do the check, but I could be missing a clever shortcut.

I think it should be worst case linear in the size of the two matchings, right? We would only need to check the number of tuples in the mapping containing each unique node ID in both first and second position, but we can do that in one pass. It's somewhat expensive, but I think it's worth it if it buys the user access to a whole lot of metrics.

PR #196, opened 2025-01-17
Update example notebooks and run in CI

### LINKED ISSUE #195: CTC example notebook errors when computing metrics
* traccuracy version: main
* Python version: 3.12
* Operating System: Ubuntu

# Description

Currently the example notebook fails for two reasons:

- The first is that the `DivisionMetrics` no longer take a tuple of frame buffer values, but an integer `max_frame_buffer` instead. This is an easy fix
- The second is that we can no longer compute `DivisionMetrics` using the `CTCMatcher` (see #194)

If we decide in #194 that we want to keep the current behaviour, we could split the example into two cells: one that computes `CTCMetrics` using `CTCMatcher`, and the other that computes `DivisionMetrics` using the `IOUMatcher`. Otherwise if we tweak the behaviour as a result of #194, we can just fix the frame buffer thing. 

# Severity
- [x] Unusable
- [ ] Very minor

msschwartz21:
Fixes #195 for now and adds notebook execution into our CI workflow.

I still think we should implement the more relaxed matcher validation discussed in #194, but this PR at least gets the notebook running and will catch any other errors that pop up while we are making a lot of changes.

# Types of Changes

- Examples

# Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -U pip
-          python -m pip install -e . nbclient
+          python -m pip install -e . nbclient ipykernel
 
       - name: Run notebooks
         # This will exit after the first error and will not report errors in subsequent nbs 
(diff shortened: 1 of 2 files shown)


cmalinmayor (approved):
LGTM!

PR #197, opened 2025-01-21
Remove unnecessary casting in get_labels_with_overlap

### LINKED ISSUE #93: [SPEED-UP] Implement `np.unique` for IoU matching
Implement fast(er) IoU computation between objects in GT and predicted mask:

```overlaps = np.unique(np.array([true, pred]), axis=0)```

### LINKED ISSUE #161: Change dtype of gt_frame from np.uint16 to np.int64
This is an edge case and a minor change.  

It is possible that for certain datasets, either the number of objects per frame are more than 65535 or that the objects are labeled non-contiguously (for example, objects are labeled as 1, 10, 100, 100000) and hence, a few object ids may take a large value. 

This could happen for a ground truth frame in practice, since often the annotation (say using a software like Mastodon) is done per cell lineage, so one would follow a cell (at t=0) and annotate all its daughters, and then return to track another cell at t=0 and follow its daughters - this might lead to non contiguous and large detection or segmentation id values for a frame in the ground truth data (in the case that there is a one-to-one mapping between the mastodon detection id and the detection id in the masks).

To handle that scenario, [this](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/4160403789bc4401e954cc28783da608efdf7df3/src/traccuracy/matchers/_compute_overlap.py#L41) could be modified  from:
```python
gt_frame = gt_frame.astype(np.uint16, copy=False)
```
to
```python
gt_frame = gt_frame.astype(np.int64, copy=False)
```
.

cmalinmayor:
# Proposed Change
Change the output type of the get_labels_with_overlap function to a list of tuples instead of a tuple of lists. Also move the validation of segmentation data types to the loaders, and just fails at TrackingGraph construction if the seg is not and integer dtype.
Fixes #161 and closes #93 as not intended (because the benchmark was 300% slower than with the numba version).

# Types of Changes
- New feature or enhancement

- Loaders
- Matchers
- Core API (Tracking Graph)

# Checklist


src/traccuracy/_tracking_graph.py
```diff
@@ -145,7 +145,17 @@ def __init__(
             name (str, optional): User specified name that will be included in result
                 outputs associated with this object
         """
-        self.segmentation = segmentation
+
+        # if the segmentation is not integer dtype, cast to uint64. Otherwise, leave it
```
msschwartz21, line 149: A statement to confirm that my understanding of the implications of this change are correct:

This could result in a situation where the type of the `label_id` on the graph is different from the type of the segmentation arrays. However, this should be fine because numpy seems to generally be sensible about comparing between floats and ints.

msschwartz21 (commented):
I'd like Draga's eyes on this too, but everything seems to check out 👍

tests/matchers/test_compute_overlap.py
```diff
@@ -19,19 +19,22 @@ def test_get_labels_with_overlap(overlap):
...
     )
 
-    perfect_gt, perfect_res, perfect_ious = get_labels_with_overlap(
-        image1, image1, overlap
-    )
-    assert list(perfect_gt) == list(range(1, n_labels + 1))
-    assert list(perfect_res) == list(range(1, n_labels + 1))
-    assert list(perfect_ious) == [1.0] * n_labels
+    ious = get_labels_with_overlap(image1, image1, overlap)
+    assert [gt for gt, _, _ in ious] == list(range(1, n_labels + 1))
```
DragaDoncila, line 23: here I think you could avoid the triple iteration by doing

```python
gt, res, iou = tuple(zip(*ious))
assert list(gt) == list(range(1, n_labels+1))
...
```

Up to you if the change is worth it

tests/matchers/test_compute_overlap.py
```diff
@@ -19,19 +19,22 @@ def test_get_labels_with_overlap(overlap):
...
+
+def test_get_labels_with_overlap_invalid():
+    n_labels = 3
+    image1 = get_annotated_image(
+        img_size=256, num_labels=n_labels, sequential=True, seed=1
     )
-    assert empty_gt == []
-    assert empty_res == []
-    assert empty_ious == []
+    with pytest.raises(ValueError):
```
DragaDoncila, line 39: I think it's good to add the match for the error message here, as otherwise the test is masking any other `ValueError`s that might be thrown that we do care about

```suggestion
    with pytest.raises(ValueError, match="Unknown overlap type: test"):
```

src/traccuracy/_tracking_graph.py
```diff
@@ -145,7 +145,17 @@ def __init__(
...
+        # the original dtype
+        if segmentation is not None and segmentation.dtype.kind not in ["i", "u"]:
+            logger.warning(
```
DragaDoncila, line 152: I think this should be a `UserWarning` not just a logger warning, because if the given arrays actually contain non-integral floats, they'll be truncated. I also think in that case, numpy will copy the array, so this could lead to memory issues for large segmentations as well.

DragaDoncila (commented):
Generally looks good, just left a couple of comments!

src/traccuracy/_tracking_graph.py
cmalinmayor: A) I didn't know about UserWarnings 😆 so thanks! B) I think in the end we should do the casting in the loaders (so that the seg_ids will match the casted data), and the tracking_graph can fail if the seg isn't integers. Two questions: 1. Do you agree that that approach makes sense? 2. Is the implmentation of the ctc loader that casts the data type if necessary as efficient as it can be? We might in-memory copy individual frames but not the whole dataset.

src/traccuracy/_tracking_graph.py
cmalinmayor: Yes. But, if we implement the casting in the loaders, the loaders can ensure that the label_id is consistent with the segmentation.

src/traccuracy/_tracking_graph.py
DragaDoncila: I think doing it in the loaders makes more sense, yeah. I think as a user I'd be surprised to see the casting happen here. I think the general approaching of casting floats is ok, as long as we tell the user about it with a `UserWarning`. Maybe also will be good to add to the warning that if you actually have non-integral values, this will result in a copy.

src/traccuracy/loaders/_ctc.py
```diff
@@ -33,11 +34,15 @@ def _load_tiffs(data_dir):
...
+    if dtype.kind not in ["i", "u"]:
+        warn(f"Segmentation has {dtype}: casting to uint64", stacklevel=2)
+        dtype = np.uint64
     stack = np.zeros(shape=shape, dtype=dtype)
-    stack[0] = first_im
+    stack[0] = first_im.astype(dtype, copy=False)
 
     for i, f in enumerate(tqdm(files[1:], "Loading TIFFs")):
-        imread(f, out=stack[i + 1])
+        stack[i + 1] = imread(f).astype(dtype, copy=False)
```
cmalinmayor, line 45: @DragaDoncila I don't think this actually results in a full copy? I guess I should do the benchmarking... I'll see if I can get it to it tomorrow.

src/traccuracy/_tracking_graph.py
```diff
@@ -147,7 +147,15 @@ def __init__(
             name (str, optional): User specified name that will be included in result
                 outputs associated with this object
         """
+
+        # if the segmentation is not integer dtype, cast to uint64. Otherwise, leave it
```
msschwartz21, line 151: Comment needs to be updated. Unfortunately I can't figure out how to do a suggestion spanning multiple lines

tests/loaders/test_ctc.py
```diff
@@ -86,3 +92,20 @@ def test_load_data_no_track_path():
...
+    test_dir = os.path.abspath(__file__)
+    data_dir = os.path.abspath(
+        os.path.join(test_dir, "../../../examples/sample-data/Fluo-N2DL-HeLa/01_RES/")
+    )
+
+    files = glob.glob(f"{data_dir}/*.tif*")
+    for file in files:
+        arr = tifffile.imread(file).astype(np.float64)
+        tifffile.imwrite(tmp_path / Path(file).name, arr)
+    with pytest.warns(UserWarning):
```
msschwartz21, line 107: Technically, we should add the `match` kwarg for the specific message

msschwartz21 (approved):
Two very minor things, otherwise LGTM!

Issue #157, opened 2024-08-26
Edge case in Mitotic Branching Correctness


lmanan:
For certain datasets, where there are no ground truth divisions and an employed tracker model also predicts no divisions, in such a scenario, the number of true positive divisions would be equal to 0, false positive divisions would be equal to 0 and false negative divisions would be equal to 0.

In this scenario, reporting the mitotic branching correctness measure equal to zero (see [here](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/4160403789bc4401e954cc28783da608efdf7df3/src/traccuracy/metrics/_divisions.py#L118)) might be misleading (?). Maybe a warning can be raised or np.nan can be returned (similar to how it is done for evaluation of F1 score [here](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html#), see documentation for `zero_division` parameter), so that it is excluded from any further evaluation.

DragaDoncila:
@lmanan thanks for raising this issue! I agree that reporting a 0 is misleading.

I would lean towards both a warning **and** a NaN, as the warning is only useful at runtime while the NaN could be written to file and persist even if the user wasn't looking at the terminal output. 

I would probably vote against making it a settable argument like they have in the F1 score for scikit-learn, because I think it's additional complexity without much benefit. @cmalinmayor @msschwartz21 curious on your thoughts?

msschwartz21:
@DragaDoncila I agree with everything you proposed 👍

msschwartz21:
Closed by #198

PR #198, opened 2025-01-21
Warn and return nans for metrics if there are no divisions in the gt

msschwartz21:
Addresses #157 by warning if there are no divisions in the ground truth and improving the handling of the subsequent zero division error. The division metrics will return np.nan except for precision which will return 0 if there are false positive divisions present.

Any other ZeroDivisionErrors will still return a metric of 0 which I think makes sense, but I would appreciate if someone double checked me on the logic with those cases.

~~Note: waiting until after #193 to merge~~

# Types of Changes

- Metrics

# Checklist

# Further Comments

### DIFF
--- a/docs/source/test_cases/divisions.pct.py
+++ b/docs/source/test_cases/divisions.pct.py
@@ -112,6 +112,9 @@ plot_matched([ex_graphs.no_children(t) for t in [0, 1]], "No Children")
 # %%
 plot_matched([ex_graphs.wrong_child(t) for t in [0, 1]], "Incorrect Child")
 
+# %%
+plot_matched([ex_graphs.wrong_children(t) for t in [0, 1]], "Incorrect Children")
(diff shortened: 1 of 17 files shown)


msschwartz21:
Technically the code coverage dropped, but I actually added better tests than we had before. The lines that aren't getting tested are the try/except blocks for ZeroDivisionErrors. With how the code is written now they really should be necessary, but I'm inclined to leave them just in case there is a weird edge case I'm not considering.  But as a consequence I can't really test them easily. Thoughts?

DragaDoncila:
@msschwartz21 I think it's fine to leave those catch-alls there. To avoid coverage failing you can explicitly add a `# pragma: no cover` comment to any line that you will purposefully not be testing

cmalinmayor (approved):
🎉 🎊

PR #200, opened 2025-01-21
Bugfix: fix testing issue with GT nodes matched to NS nodes

### LINKED ISSUE #191: Fix ns vertex bug with false negative flag
https://github.com/Janelia-Trackathon-2023/traccuracy/blob/b367f32072a7c0eb9c4dab9ba6ee50aeebaa7819/tests/track_errors/test_ctc_errors.py#L277-L280

This bug is documented in the code, but I keep forgetting about it since it doesn't have an open issue.

msschwartz21:
Fixes #191 

When I tried, I couldn't reproduce the original issue that led us to comment out that block in the test. All I actually did here was update the code to reflect that GT nodes matched to a NS node are not annotated with any error type including FN.

# Types of Changes

- Track Errors

# Checklist

### DIFF
--- a/tests/track_errors/test_ctc_errors.py
+++ b/tests/track_errors/test_ctc_errors.py
@@ -273,10 +273,8 @@ def test_ns_vertex_fn_edge():
     for edge in comp_edges:
         assert EdgeFlag.CTC_FALSE_POS not in comp.edges[edge]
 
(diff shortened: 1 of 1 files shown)


DragaDoncila (approved):
I think this is correct based on the metric. When multiple GT nodes are matched to a single NS nodes, the GT nodes don't get any kind of flag, the computed nodes matched to that GT node get the NS flag.

PR #202, opened 2025-01-29
Compute the matching type for each dataset and use for metric validation

msschwartz21:
Closes #170 by switching over to a `Matched` property that returns the type of matching. Updates the Metric class and subclasses to use this information instead of the previous logic based on the exact name of the Matcher.

# Types of Changes
- New feature or enhancement

- Matchers
- Metrics

# Checklist

### DIFF
--- a/src/traccuracy/matchers/_base.py
+++ b/src/traccuracy/matchers/_base.py
@@ -139,6 +139,7 @@ class Matched:
             pred_type = "one"
 
         self._matching_type = f"{gt_type}-to-{pred_type}"
+        self.matcher_info["matching type"] = self._matching_type
         return self._matching_type
 
     def _get_match(self, node: Hashable, map: dict[Hashable, list]):
--- a/src/traccuracy/matchers/_base.py
+++ b/src/traccuracy/matchers/_base.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
 import logging
+import warnings
 from abc import ABC, abstractmethod
-from collections import defaultdict
+from collections import Counter, defaultdict
 from typing import Any, Hashable
 
 from traccuracy._tracking_graph import TrackingGraph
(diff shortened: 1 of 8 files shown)


src/traccuracy/matchers/_base.py
```diff
@@ -111,6 +112,36 @@ def __init__(
...
+        """Determines the matching type: one-to-one, one-to-many, many-to-one, many-to-many"""
+        if len(self.mapping) == 0:
+            warnings.warn(
+                "Mapping is empty. Defaulting to type of one-to-one", stacklevel=2
+            )
+
+        if self._matching_type is not None:
+            return self._matching_type
+
+        gt_count = Counter(m[0] for m in self.mapping)
```
DragaDoncila, line 128: I think we could get away with a slightly cheaper computation here by iterating through the `gt_pred_map` and `pred_gt_map` directly, and breaking when we hit a duplicate. Something like

```python
gt_type = "one"
for label_id, matches in self.gt_pred_map.items():
   if len(matches) > 1:
       gt_type = "many"
       break
```

And then the same for the `pred_type`. In the worst case we still make one full iteration of each, but we save on creating the `Counter` object and then casting to `set`

tests/test_run_metrics.py
```diff
@@ -41,24 +37,25 @@ def test_run_metrics():
...
-        run_metrics(graph, graph, DummyMatcher(), [DummyMetric])
+    # # Check metric input -- not instantiated
+    # with pytest.raises(TypeError):
+    #     run_metrics(graph, graph, DummyMatcher(), [DummyMetric])
 
-    # Check metric input -- wrong type
-    with pytest.raises(TypeError):
-        run_metrics(graph, graph, DummyMatcher(), [DummyMetric(), "rando"])
+    # # Check metric input -- wrong type
+    # with pytest.raises(TypeError):
```
DragaDoncila, line 52: Curious why we've commented these out, was that just testing purposes?

DragaDoncila (commented):
Looks great, left a comment with a slight improvement for the match type checking. Would be good to also make sure the `override_matcher` option is tested

src/traccuracy/matchers/_base.py
```diff
@@ -111,6 +112,36 @@ def __init__(
         self.gt_pred_map = dict(gt_pred_map)
         self.pred_gt_map = dict(pred_gt_map)
 
+        self._matching_type = None
+
+    @property
+    def matching_type(self):
+        """Determines the matching type: one-to-one, one-to-many, many-to-one, many-to-many"""
```
cmalinmayor, line 119: Can we document which is first, gt or pred?

tests/test_run_metrics.py
msschwartz21: Yup my bad. Just from debugging something

src/traccuracy/matchers/_base.py
msschwartz21: That's much more elegant!

cmalinmayor (approved):
Looks good now!

PR #206, opened 2025-02-13
Make consistent behavior when denom is 0 and add docstrings

cmalinmayor:
@DragaDoncila I was going to just add docstrings and type annotations, and then I noticed that the behavior was inconsistent between the different metrics for what happens when the denomintor is 0.

In this PR I change them all to warn and return np.nan, whereas previously the TRA raised a RuntimeError, the LNK warned and returned np.nan, and the DET did not check. I don't think I have strong opinions about which of these options is better, other than that checking for a zero denominator is good, so this is just a suggestion of one possible solution.

I didn't fix the tests, since I wanted your opinion on what the "correct" solution is first. Feel free to just pull the bits of code you want and treat this as a review rather than a PR - I just already wrote the code so I figured I would provide it.

### DIFF
--- a/src/traccuracy/metrics/_ctc.py
+++ b/src/traccuracy/metrics/_ctc.py
@@ -117,9 +117,6 @@ class CTCMetrics(AOGMMetrics):
             n_nodes (int): The number of nodes in the ground truth graph
             n_edges (int): The number of edges in the ground truth graph
 
-        Raises:
-            RuntimeError: if the AOGM0 is 0
-
         Returns:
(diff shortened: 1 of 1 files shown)


DragaDoncila:
@cmalinmayor yep thanks for flagging. I initially made the TRA an error as you would be unable to compute any of the CTC metrics downstream, but the user may have requested other metrics as well, so I think going consistently with a `nan` and warning makes more sense! I'm going to merge this PR into the branch and fix tests etc.

PR #218, opened 2025-02-26
Improve runtime of `TrackOverlap` metrics

### LINKED ISSUE #132: Improve runtime performance for track overlap benchmarking
In the initial work on #117, we discovered that the runtime on track overlap metrics was sufficiently long that it was causing issues with running the benchmarking on github actions. @cmalinmayor started to work on optimizations for the function, but I just want to open this issue so that we don't forget about it. 

# Severity
- [x] Unusable
- [ ] Very minor

DragaDoncila:
# Proposed Change
Closes #132.

Improve runtime performance of `TrackOverlap` metrics by leveraging our already computed node matching to avoid checking completely unrelated tracklets.

Prior to this PR, we were maintaining a list of all possibly overlapping mapped edges, and using a set intersect on each reference tracklet to get the actual number of overlapping edges.

In this PR, I use the knowledge that only edges in a reference tracklet with **both endpoints present in the node matching** could possibly result in an overlapping edge. Rather than keeping a list of all possibly mapped edges, we immediately check whether a proposed overlapping edge of the reference tracklet is present in any of the overlap tracklets (this check is O(1) because I build a dictionary mapping `edge -> tracklet id` in a linear pass over the tracklets). If the edge is present, we increment the overlap count of the given tracklet ID for that reference tracklet. After we've checked all edges of the reference tracklet, I get the maximum count of all tracklet IDs that have any overlap with the reference tracklet.

Since the metric supports `many-to-one` matching, an unhappy path for this metric remains, when many overlap nodes are mapped to both the source and target of a given reference edge e.g. for reference edge `uv` we have 10 nodes mapped to `u` and 10 nodes mapped to `v`. In this case we'd be checking 100 "possible" edges for presence in the overlap graph. I think this is highly improbable. We will most of the time have `one-to-one` mapping, or maybe a handful of nodes mapped.

# Types of Changes
- Enhancement

- Track Overlap Metrics

# Checklist


DragaDoncila:
@msschwartz21 @cmalinmayor it would be great if you could check my reasoning here. I would also like to actually get some benchmarking going, but keen to hear your thoughts first. To implement benchmarking I was just going to copy what Caroline does in #209 haha.

DragaDoncila:
Also I will need to explicitly write more tests with a `many-to-one` matching involved, I think.

msschwartz21:
Re benchmarking, I started working on something in #117, but didn't get very far because I ended up just trying to figure out why the action was failing. But you could use it as a starting point. Feel free to just copy over what you need because the branch is super stale.

DragaDoncila:
The failing benchmark is due to a memory issue, because we copy the graph and then further use memory for the tracklet subgraph and the overlap dictionary. I think it can be addressed by improving the `get_tracklets` function so I will take a look at that

cmalinmayor (approved):
I opened #225 for adding docs and standard test cases, since I started to do it here and it was non-trivial. I did run the overlap metrics on one of my experiments and it worked quickly, so this looks good to me!

DragaDoncila:
@cmalinmayor @msschwartz21 usually I don't self-merge PRs but is that our workflow after approval, and if so, is this one ready to go?

msschwartz21:
Yes that is what Caroline and I have been doing. Go ahead and merge if you are happy with it :)

Issue #216, opened 2025-02-24
Drop python 3.9


cmalinmayor:

msschwartz21:
- [ ] Remember to bump the ruff target version to py310

Issue #215, opened 2025-02-24
Unify running python in docs


cmalinmayor:
@DragaDoncila suggested myst markdown - given an example, @msschwartz21 will migrate our existing notebook formats to this (or somehow unify the format into one).

DragaDoncila:
Ok so one example of a MyST Markdown notebook is [this napari docs page](https://napari.org/stable/developers/contributing/documentation/docs_template.html), and this is the [source file ](https://github.com/napari/docs/blob/main/docs/developers/contributing/documentation/docs_template.md). If you look at the [raw file](https://raw.githubusercontent.com/napari/docs/refs/heads/main/docs/developers/contributing/documentation/docs_template.md), you'll see it's mostly markdown with a couple of code cells, and the only necessary config is the preamble `jupytext` block at the top. There's also a nice VSCode extension for it.

Looking at the [getting started](https://myst-nb.readthedocs.io/en/v0.13.2/use/start.html) guide, we should only need to add the `myst-nb` sphinx extension to `conf.py` and our docs requirements. [Here's](https://myst-nb.readthedocs.io/en/v0.13.2/use/markdown.html) a bit more about MyST notebooks specifically, and it looks like [there's a CLI](https://rst-to-myst.readthedocs.io/en/latest/) to convert between rst and md but I haven't played with it yet.

Let me know if this is enough to go on @msschwartz21, I am no docs expert myself.

msschwartz21:
@DragaDoncila Thank you! That's more than enough to get me started and it looks like this is definitely the right solution for the various use cases that led to the current medley of formats :joy:

msschwartz21:
I've gotten started on the conversion although I'm going to wait until #220 is merged in so that I can also add the pixi tasks in the same PR. A few questions have come up:

- Currently we have the ctc example notebook stored in `examples/ctc.ipynb` which is then linked and rendered in the documentation using `nbsphinx` and `nbsphinx-link`. We could choose to convert the notebook to mystmd and let it live directly in the documentation. Pros: eliminates the need for `nbsphinx` and `nbsphinx-link`. Cons: the ipynb isn't directly available if someone wants to run it, although I think we should be able to add a "download as ipynb" button to the docs.
- If we make this change, where would the sample data live? Currently it's in `examples/sample-data`, but that wouldn't make as much sense if we move the notebook.

Issue #213, opened 2025-02-24
Developer docs update


msschwartz21:
- [ ] Running benchmarking locally (after downloading testing data)
- [ ] Building the docs locally

msschwartz21:
The docs instructions need to be updated after incorporating pixi

PR #220, opened 2025-03-04
Configure pixi for traccuracy

msschwartz21:
This PR expands pyproject.toml with configuration for `pixi` which we can use for environment management during development. All of the normal pip-based installation options will still work, but `pixi` adds an additional layer of functionality. Right now pixi will use python 3.10 as the default, but I will probably bump this to the latest version that we support when we make python changes in #216 and #207. 

Four tasks are currently available through pixi
- benchmark -> run benchmarking locally
- getdata -> download test data used for benchmarking
- test-case-report -> generate test case coverage report
- test -> run normal tests

I've updated the developer docs #213 to describe how to use pixi for normal development tasks. I will add tasks for building documentation after in the PR to address #215.

# Types of Changes

# Checklist


cmalinmayor (approved):
After extensive in-person review: I approve! 🌮

PR #226, opened 2025-03-11
Update test case report to contain the new group for gap closing

msschwartz21:


DragaDoncila:
Thanks for adding this @msschwartz21!

PR #236, opened 2025-03-25
CTC loader checks dimensionality of data to set TrackingGraph loc keys

### LINKED ISSUE #235: Loading 3D data with ctc loader silently drops z dimension
When we load 3D data with the ctc loader, example from benchmarking
```python
def gt_data_3d():
    path = "downloads/Fluo-N3DH-CE/01_GT/TRA"
    return load_ctc_data(
        os.path.join(ROOT_DIR, path),
        os.path.join(ROOT_DIR, path, "man_track.txt"),
        run_checks=False,
    )
```

If you retrieve node locations from the resulting `TrackGraph`, the node locations only contain x and y. This is because the `TrackingGraph` defaults to location keys `('x', 'y')`, even the nodes contain a z attribute. 

At the very least, we should update `load_ctc_data` to check if the data is 3D and set the location keys accordingly.

For the `TrackingGraph`, I don't think we want to try to make inferences about dimensionality and that should be left to the discretion of the user.

msschwartz21:
Closes #235 

# Types of Changes

- Loaders

# Checklist

### DIFF
--- a/src/traccuracy/loaders/_ctc.py
+++ b/src/traccuracy/loaders/_ctc.py
@@ -269,4 +269,11 @@ def load_ctc_data(
             "Consider setting `run_checks=True` for detailed error message."
         ) from e
 
-    return TrackingGraph(G, segmentation=masks, name=name)
+    # Check dimensionality of segmentation data to set the location keys
+    loc_keys: tuple[str, ...]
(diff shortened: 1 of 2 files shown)


cmalinmayor (approved):
🎋

PR #237, opened 2025-04-01
Allow positions stored as lists in a single attribute

cmalinmayor:
# Proposed Change
Expands set of valid location keys to include a single location key holding a list of values, along with the existing tuple of keys holding one value each.

# Types of Changes
What types of changes does your code introduce?
- New feature or enhancement
- Tests and benchmarks

Which topics does your change affect? 

# Checklist

# Further Comments
We could rename the argument to "location_key", but that is a breaking change.
We could also pick one type of location key instead of supporting both formats, but that is another big breaking change. I've found needing to rewrite all the attributes on my nodes to be a pretty big usability barrier, and supporting both is a small code burden, so I'm in favor of keeping both options.

### DIFF
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -118,7 +118,7 @@ class TrackingGraph:
             The name of the node attribute that corresponds to the frame of
             the node. Defaults to "t".
         location_keys: tuple of str | str
-            Keys used to access the location of the cell in space.
+            Key(s) used to access the location of the cell in space.
     """
 
     def __init__(
@@ -153,9 +153,10 @@ class TrackingGraph:
                 'segmentation_id'. Pass `None` if there is not a label
                 attribute on the graph.
             location_key (str | tuple, optional): The key or list of keys on each node
-                in graph that contains the spatial location of the node. Every
+                in graph that contains the spatial location of the node, in the order
+                needed to index the segmentation, if applicable. Every
                 node must have a value stored at each of these provided keys.
-                If a single string, it is assume that the location is stored as a list
+                If a single string, it is assumed that the location is stored as a list
                 on each node in a single attribute. Defaults to ('x', 'y').
             name (str, optional): User specified name that will be included in result
                 outputs associated with this object
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -65,6 +65,24 @@ class NodeFlag(str, enum.Enum):
         return value in cls.__members__.values()
 
 
+def _check_valid_key_name(key: str, name: str) -> None:
+    """Check if the provided key conflicts with the reserved NodeFlag values and raise
+    a ValueError if so.
+
+    Args:
+        key (str): The key to check if it conflicts with the reserved values
+        name (str): The name of key to use in the error message
+
+    Raises:
+        ValueError: if the provided key conflicts with the NodeFlag values
+    """
+    if NodeFlag.has_value(key):
+        raise ValueError(
+            f"Specified {name} key {key} is reserved for graph"
+            f"annotation. Please change the {name} key."
+        )
+
+
 @enum.unique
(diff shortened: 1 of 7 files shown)


src/traccuracy/_tracking_graph.py
```diff
@@ -152,10 +152,11 @@ def __init__(
                 pixel value of the node in the segmentation. Defaults to
                 'segmentation_id'. Pass `None` if there is not a label
                 attribute on the graph.
-            location_keys (tuple, optional): The list of keys on each node in
-                graph that contains the spatial location of the node. Every
-                node must have a value stored at each of these keys.
-                Defaults to ('x', 'y').
+            location_key (str | tuple, optional): The key or list of keys on each node
```
DragaDoncila, line 155:
```suggestion
            location_key (str | tuple, optional): The key or tuple of keys on each node
```

src/traccuracy/_tracking_graph.py
```diff
@@ -152,10 +152,11 @@ def __init__(
...
+                in graph that contains the spatial location of the node. Every
+                node must have a value stored at each of these provided keys.
+                If a single string, it is assume that the location is stored as a list
```
DragaDoncila, line 158:
```suggestion
                If a single string, it is assumed that the location is stored as a list
```

DragaDoncila: Does the location key, if a single str, have to point to a `list` specifically? Can it point to a tuple? Is `np.ndarray` ok? I think we should be specific here, otherwise say `iterable`? I don't love forcing it to strictly be a list, as `np.ndarray` is pretty common, at a minimum

src/traccuracy/_tracking_graph.py
```diff
@@ -169,18 +170,23 @@ def __init__(
...
-        if label_key is not None and NodeFlag.has_value(label_key):
-            raise ValueError(
-                f"Specified label key {label_key} is reserved for graph"
-                "annotation. Please change the label key."
-            )
-        self.label_key = label_key
-        for loc_key in location_keys:
-            if NodeFlag.has_value(loc_key):
+
+        def _check_valid_key_name(key: str, name: str) -> None:
```
DragaDoncila, line 174: Would just pull this function outside of the class, either in the same file or in a `utils` type file.

DragaDoncila (commented):
Looks good, just left some comments for clarification. I kinda agree with Morgan that maybe we don't provide a default value

cmalinmayor:
Crazy idea: should we make the location attribute optional, if there is a segmentation provided? The segmentation-based matchers don't use the location attribute.

src/traccuracy/_tracking_graph.py
cmalinmayor: I was trying to annotate the return type of `get_location` to be consistent with your suggestion. I tried Iterable, but we call 'len' on it. I tried Collection, but we index into it. I am settling for a type annotation of `list | tuple | np.ndarray`, I guess? Or I can cast to list when I return from get_location, but that feels wasteful.

DragaDoncila: >  I am settling for a type annotation of list | tuple | np.ndarray

I think that's fine! Explicit is better than implicit. Would be good to test that `np.ndarray` works ok but I don't think it's strictly necessary for this PR

DragaDoncila:
> Crazy idea: should we make the location attribute optional, if there is a segmentation provided? The segmentation-based matchers don't use the location attribute.

In theory this should work just fine, but I fear it :laughing:

cmalinmayor:
Conclusion from meeting: Default location key is None, don't require it on the graph, provide useful error message if Point matching and no location key provided.

cmalinmayor:
@DragaDoncila Can you (re)review if you get the chance?

DragaDoncila:
@cmalinmayor added it to my todos for tomorrow!

DragaDoncila (approved):
Looks good to me @cmalinmayor!

PR #243, opened 2025-04-21
Sort nodes to ensure point matcher is deterministic

### LINKED ISSUE #240: Non-deterministic AOGM values with PointMatcher
I've observed that repeated runs using the `PointMatcher` sometimes yield slightly different AOGM values, even with the same input data. This non-determinism appears to stem from differences in node ordering when constructing the cost matrix.

Specifically, replacing this code [here](https://github.com/Janelia-Trackathon-2023/traccuracy/blob/5fd8fc55196bec2785e5220976afd75b6746b650/src/traccuracy/matchers/_point.py#L52):

```python
 gt_nodes = list(gt_graph.nodes_by_frame.get(frame, []))
 pred_nodes = list(pred_graph.nodes_by_frame.get(frame, []))
```
with

```python
gt_nodes = sorted(gt_graph.nodes_by_frame.get(frame, []))
pred_nodes = sorted(pred_graph.nodes_by_frame.get(frame, []))
```
eliminated the variation across runs.

I am guessing that in frames where there are multiple equally optimal matchings (i.e., ties in the cost matrix), the `linear_sum_assignment` solver may return different solutions depending on the input order. Sorting the nodes ensures a consistent matrix layout, which in turn produces deterministic matchings.

cmalinmayor:
# Proposed Change
Closes #240 

# Types of Changes

- Matchers

# Checklist

### DIFF
--- a/src/traccuracy/matchers/_point.py
+++ b/src/traccuracy/matchers/_point.py
@@ -49,9 +49,12 @@ class PointMatcher(Matcher):
         if gt_graph.start_frame is None or gt_graph.end_frame is None:
             return mapping
         for frame in range(gt_graph.start_frame, gt_graph.end_frame):
+            # Sorting node ids to ensure deterministic solution when there are ties
+            # Ignoring typing because technically "Hashable" node ids are not
+            # always sortable, but we don't anticipate non-sortable types
             gt_nodes = sorted(gt_graph.nodes_by_frame.get(frame, []))  # type: ignore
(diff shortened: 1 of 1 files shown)


cmalinmayor:
@msschwartz21 mypy is unhappy with sorting a list of hashable node ids. Are you okay with ignoring this particular mypy error, on the assumption that while technically someone could make node ids that are not sortable but are hashable, they do so at their own peril?

msschwartz21 (approved):

PR #244, opened 2025-04-25
Load csv into TrackingGraph

### LINKED ISSUE #219: Point data loader
# Description

To complete the point based feature set started in #209, we should add a loader for point data that does not depend on segmentations.

# Topics

- New feature or enhancement

- Loaders

# Priority
- [ ] This is an essential feature
- [x] Nice to have
- [ ] Future idea

- [x] Yes! :tada:

msschwartz21:
# Proposed Change
Closes #219 

Currently I'm assuming that the csv-like file contains node positions (t, z, y, x) and a "parent" column that specifies the node id of the parent node in the previous timepoint. If there is another common way of dumping to a csv where the edges are encoded differently, lmk and I can add support for it.

# Types of Changes
- New feature or enhancement

- Loaders

# Checklist

# Further Comments

### DIFF
--- a/src/traccuracy/loaders/_point.py
+++ b/src/traccuracy/loaders/_point.py
@@ -1,4 +1,4 @@
-from typing import cast
+from typing import Optional, cast
 
 import networkx as nx
 import pandas as pd
@@ -8,7 +8,7 @@ from traccuracy._tracking_graph import TrackingGraph
 
 def load_point_data(
     path: str | None = None,
-    df: pd.DataFrame | None = None,
+    df: Optional[pd.DataFrame] = None,  # noqa: UP007, bar syntax breaks docks build
     parent_column: str = "parent",
     id_column: str | None = None,
     pos_columns: tuple[str, ...] = ("z", "y", "x"),
--- a/src/traccuracy/loaders/_point.py
+++ b/src/traccuracy/loaders/_point.py
@@ -9,7 +9,7 @@ from traccuracy._tracking_graph import TrackingGraph
 
 def load_point_data(
     path: str | None = None,
-    df: Optional[pd.DataFrame] = None,  # noqa: UP007, bar syntax breaks docks build
+    df: Optional[pd.DataFrame] = None,  # noqa: UP007, bar syntax breaks docs build
     parent_column: str = "parent",
     id_column: str = "node_id",
     pos_columns: tuple[str, ...] = ("z", "y", "x"),
--- a/src/traccuracy/loaders/_point.py
+++ b/src/traccuracy/loaders/_point.py
@@ -10,11 +10,13 @@ def load_point_data(
     path: str | None = None,
     df: Optional[pd.DataFrame] = None,  # noqa: UP007, bar syntax breaks docks build
     parent_column: str = "parent",
-    id_column: str | None = None,
+    id_column: str = "node_id",
     pos_columns: tuple[str, ...] = ("z", "y", "x"),
     time_column: str = "t",
+    seg_id_column: str | None = None,
     name: str | None = None,
     sep: str | None = None,
+    no_parent_val: int | None = -1,
 ) -> TrackingGraph:
     """Load point-based tracking data into a TrackingGraph from a csv-like file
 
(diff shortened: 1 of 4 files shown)


src/traccuracy/loaders/_point.py
```diff
@@ -0,0 +1,90 @@
...
+    - position, e.g. three columns 'z', 'y', 'x'
+    - parent, a reference to the node in the previous time frame
+
+    Args:
+        path (str | None, optional): Path to the csv-like file to load. Defaults to None.
+        df (pd.DataFrame | None, optional): A dataframe that has already been loaded.
+            Defaults to None.
+        parent_column (str | None, optional): A reference to the parent node in the previous
+            time frame. Defaults to "parent".
+        id_column (str | None, optional): Optional column used to specify node ids.
```
cmalinmayor, line 32: I don't think the default being None makes sense combined with "parent." Whoever wrote out the parent ID would have to know the default that you picked (row index) and there is a big risk of mess if that's not right. I think we should require a node id column.

src/traccuracy/loaders/_point.py
```diff
@@ -0,0 +1,90 @@
...
+            Defaults to None and the row index is used as the id
+        pos_columns (tuple[str], optional): A tuple of columns to use for position.
+            Defaults to ("z", "y", "x").
+        time_column (str, optional): The column to use for time. Defaults to "t".
+        name (str | None, optional): Optional string to name/describe the dataset. Defaults to None.
+        sep (str | None, optional): Passed to pd.read_csv to set the sep kwarg. Defaults to None.
+
```
cmalinmayor, line 39: Are we envisioning this only used when segmentations aren't present, or additionally when you have a segmentation and have already done your regionprops-like thing to create the graph? I think there is a use case for saving the graph as a CSV alongside the segmentation to avoid the need to re-compute the graph every time you do evaluation. In which case we should have a seg_id column (or whatever the segmentation loaders use to match segs to nodes)

src/traccuracy/loaders/_point.py
```diff
@@ -0,0 +1,90 @@
...
+    path: str | None = None,
+    df: Optional[pd.DataFrame] = None,  # noqa: UP007, bar syntax breaks docks build
+    parent_column: str = "parent",
+    id_column: str | None = None,
+    pos_columns: tuple[str, ...] = ("z", "y", "x"),
+    time_column: str = "t",
+    name: str | None = None,
+    sep: str | None = None,
+) -> TrackingGraph:
+    """Load point-based tracking data into a TrackingGraph from a csv-like file
```
cmalinmayor, line 19: Add to the docs that any additional columns are added as node attributes (they are, right?)

src/traccuracy/loaders/_point.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Assumes each row contains:
+    - time
+    - position, e.g. three columns 'z', 'y', 'x'
+    - parent, a reference to the node in the previous time frame
```
cmalinmayor, line 24: Document how to say there is no parent (-1? None? either?)

cmalinmayor (changes requested):
Mostly documentation/design questions - implementation looks good!

src/traccuracy/loaders/_point.py
msschwartz21: I decided not to add any additional columns since traccuracy wouldn't use the, but I could add a flag in to make it optional in case someone wants it for visualization downstream.

src/traccuracy/loaders/_point.py
msschwartz21: Do you think I should make this a settable option?

src/traccuracy/loaders/_point.py
```diff
@@ -10,11 +10,13 @@ def load_point_data(
     path: str | None = None,
     df: Optional[pd.DataFrame] = None,  # noqa: UP007, bar syntax breaks docks build
```
cmalinmayor, line 11: docks ⛵

src/traccuracy/loaders/_point.py
cmalinmayor: I think it makes sense to ignore them if traccuracy doesn't use them - if people really want them they can make a custom loader.

src/traccuracy/loaders/_point.py
```diff
@@ -29,20 +31,25 @@ def load_point_data(
...
+        id_column (str, optional): Optional column used to specify node ids.
+            Defaults to "node_id"
         pos_columns (tuple[str], optional): A tuple of columns to use for position.
             Defaults to ("z", "y", "x").
         time_column (str, optional): The column to use for time. Defaults to "t".
+        seg_id_column (str | None, optional): Name of an optional column containing a segmentation
+            label id. Defaults to None.
         name (str | None, optional): Optional string to name/describe the dataset. Defaults to None.
         sep (str | None, optional): Passed to pd.read_csv to set the sep kwarg. Defaults to None.
+        no_parent_val (str | None, optional): The value used to indicate that a node
```
cmalinmayor, line 43: I don't think you need to make this an argument. I think you can just document what the CSV should look like (I prefer -1 OR None, but up to you).

PR #245, opened 2025-04-28
Remove graph copying in division metrics with frame buffer

DragaDoncila:
If you are implementing a new matcher or metric, please append this `&template=new_matcher_metric.md` to your url to load the correct template.

# Proposed Change

This PR removes the need to generate graph copies for each frame buffer value being tested in calculating division metrics.

Annotated graphs will now retain their `FP_DIV`/`FN_DIV` annotations for divisions incorrect at `frame_buffer=0` and gain an additional attribute `min_buffer_correct` that stores `np.nan` when a buffer value that corrects the division hasn't been tried, otherwise the minimal frame buffer that would correct the division.

Note that in this implementation, **divisions will remain correct for buffer values greater than the minimal buffer value that corrects the division**.

This PR also adds a utility function that returns copies of the annotated graphs with all divisions corrected by the given frame buffer value marked as `TP_DIV`.

# Types of Changes
- New feature or enhancement
- Documentation update

- Metrics

# Checklist

### DIFF
--- a/src/traccuracy/metrics/_divisions.py
+++ b/src/traccuracy/metrics/_divisions.py
@@ -86,19 +86,12 @@ class DivisionMetrics(Metric):
         Returns:
             dict: Returns a nested dictionary with one dictionary per frame buffer value
         """
-        data = _evaluate_division_events(
+        _evaluate_division_events(
             data,
             max_frame_buffer=self.frame_buffer,
         )
 
         return self._calculate_metrics(data.gt_graph, data.pred_graph)
-        # return {
-        #     f"Frame Buffer {fb}": self._calculate_metrics(
-        #         matched_data.gt_graph,
-        #         matched_data.pred_graph,
-        #     )
-        #     for fb, matched_data in div_annotations.items()
-        # }
 
     def _get_mbc(self, gt_div_count: int, tp_division_count: int, fp_division_count: int) -> float:
         """Computes Mitotic Branching Correctness and returns nan if there are no gt
@@ -116,7 +109,9 @@ class DivisionMetrics(Metric):
             return np.nan
         return tp_division_count / (fp_division_count + gt_div_count)
 
-    def _calculate_metrics(self, g_gt: TrackingGraph, g_pred: TrackingGraph) -> dict[str, float]:
+    def _calculate_metrics(
+        self, g_gt: TrackingGraph, g_pred: TrackingGraph
+    ) -> dict[str, dict[str, float]]:
         gt_div_count = len(g_gt.get_divisions())
         pred_div_count = len(g_pred.get_divisions())
         tp_division_count = len(g_gt.get_nodes_with_flag(NodeFlag.TP_DIV))
--- a/src/traccuracy/metrics/_divisions.py
+++ b/src/traccuracy/metrics/_divisions.py
@@ -86,11 +86,12 @@ class DivisionMetrics(Metric):
         Returns:
             dict: Returns a nested dictionary with one dictionary per frame buffer value
         """
-        div_annotations = _evaluate_division_events(
+        data = _evaluate_division_events(
             data,
             max_frame_buffer=self.frame_buffer,
         )
 
+        return self._calculate_metrics(data.gt_graph, data.pred_graph)
         return {
             f"Frame Buffer {fb}": self._calculate_metrics(
                 matched_data.gt_graph,
@@ -131,7 +132,8 @@ class DivisionMetrics(Metric):
         f1 = self._get_f1(recall, precision)
         mbc = self._get_mbc(gt_div_count, tp_division_count, fp_division_count)
 
-        return {
+        res_dict = {}
+        res_dict["Frame Buffer 0"] = {
             "Division Recall": recall,
             "Division Precision": precision,
             "Division F1": f1,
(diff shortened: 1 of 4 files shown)


src/traccuracy/utils.py
```diff
@@ -0,0 +1,37 @@
+import copy
+
+import numpy as np
+
+from traccuracy import NodeFlag, TrackingGraph
+from traccuracy.matchers import Matched
```
DragaDoncila, line 6: I tried to put this under type checking but then tests failed. I don't understand why this isn't considered purely a typing use...

DragaDoncila:
@msschwartz21 [the warning](https://app.readthedocs.org/projects/traccuracy/builds/27988948/#269028301--188) we get for docs is that there's multiple targets for `TrackingGraph` and `Matched`... is it something to do with my import in `utils`...? I have no idea :grimacing:

src/traccuracy/utils.py
msschwartz21, line 6: I'm not sure about the typing issue, but for the docs problem can you try switching the imports like below and see if that fixes the problem?
```suggestion
from traccuracy._tracking_graph import NodeFlag, TrackingGraph
from traccuracy.matchers._base import Matched
```

msschwartz21: @cmalinmayor can probably double check me on this, but would adding `from __future__ import annotations` allow you put those imports under type checking?

src/traccuracy/utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+import numpy as np
+
+from traccuracy._tracking_graph import NodeFlag, TrackingGraph
+
+if TYPE_CHECKING:
+    from traccuracy.matchers._base import Matched
+
+
+def get_corrected_division_graphs_with_delta(
```
msschwartz21, line 14: Just checking, this is intended as a utility function in case someone wants to export the graph for visualization or some other task with the annotations for a particular frame buffer.

src/traccuracy/track_errors/divisions.py
```diff
@@ -165,8 +164,17 @@ def _correct_shifted_divisions(matched_data: Matched, n_frames: int = 1) -> Matc
...
-        t_fn = g_gt.graph.nodes[fn_node][g_gt.frame_key]
+        fp_node_info = g_pred.graph.nodes[fp_node]
+        fn_node_info = g_gt.graph.nodes[fn_node]
+        t_fp = fp_node_info[g_pred.frame_key]
+        t_fn = fn_node_info[g_gt.frame_key]
+
+        # Move on if this division has already been corrected by a smaller buffer value
+        if (
+            fp_node_info.get("min_buffer_correct", np.nan) is not np.nan
+            and fn_node_info.get("min_buffer_correct", np.nan) is not np.nan
```
msschwartz21, line 175: Should this be an or? E.g. if one of them has a non nan value that means that it's already been matched to another division that can be corrected with a frame shift.

src/traccuracy/metrics/_divisions.py
```diff
@@ -143,3 +140,31 @@ def _calculate_metrics(self, g_gt: TrackingGraph, g_pred: TrackingGraph) -> dict
             "False Negative Divisions": fn_division_count,
             "Wrong Children Divisions": wc_division_count,
         }
+        for fb in range(1, self.frame_buffer + 1):
+            new_tp_div_count = 0
+            for node in g_pred.graph.nodes:
```
msschwartz21, line 145: Can this loop be shortened by just iterating over the division nodes?

src/traccuracy/utils.py
```diff
@@ -0,0 +1,43 @@
...
+    Args:
+        matched (traccuracy.matchers.Matched): Matched object for set of GT and Pred data.
+            Must be annotated with division events.
+        frame_buffer (int): Maximum frame buffer to use for division correction
+
+    Returns:
+        tuple[traccuracy.TrackingGraph, traccuracy.TrackingGraph]: Tuple of corrected
+        GT and Pred graphs
+    """
+    corrected_gt_graph = copy.deepcopy(matched.gt_graph)
```
msschwartz21, line 31: We can check the `matched.gt_graph.division_annotations` flag to check that the divisions have been annotated.

msschwartz21 (commented):
Just realized that you hadn't rerequested review yet, so this may be a bit early on my part. Code all looks good to me except for a couple minor comments.

It looks like there are still some issues with the track errors docs. Let me know if I can help figure that out.

msschwartz21 (commented):
Just realized that you hadn't rerequested review yet, so this may be a bit early on my part. Code all looks good to me except for a couple minor comments.

It looks like there are still some issues with the track errors docs. Let me know if I can help figure that out.

DragaDoncila:
Add reserved keyword for "min_buffer_correct"

DragaDoncila:
> It looks like there are still some issues with the track errors docs. Let me know if I can help figure that out.

@msschwartz21 I know we looked at this during the meeting but I forgot which pages were wrong and comparing them now with the live docs they all look identical :sob: Can you please remind me again what the issue was...

msschwartz21:
I just took a look at the current docs and the track errors API and it is already messed up in the current version of the docs. I'll open an issue and add it to the list of docs maintenance.

src/traccuracy/track_errors/divisions.py
DragaDoncila: Yep, good catch!

src/traccuracy/metrics/_divisions.py
DragaDoncila: This is a good question, and I think it's worth us discussing in other contexts. I'm not sure. `TrackingGraph.get_divisions()` technically runs a loop of length `number_of_nodes`, because it gets the `out_degree` and then filters the nodes by degree. So it could actually be slower to `get_divisions` + iterate through divisions than to just iterate through all nodes?

DragaDoncila: I think in this case it would probably still be quicker, since we won't be getting the node info for unrelated nodes, so I've updated it. I think it would be a good idea to consider more broadly whether some of our convenience functions are inadvertently increasing complexity. Maybe we should store dividing nodes on graph build so that we can access them without iterating?

src/traccuracy/_tracking_graph.py
```diff
@@ -49,6 +49,8 @@ class NodeFlag(str, enum.Enum):
     FALSE_POS = "is_fp"
     FALSE_NEG = "is_fn"
 
+    MIN_BUFFER_CORRECT = "min_buffer_correct"
```
DragaDoncila, line 52: I realized adding this here also means we keep a set for it in the `TrackingGraph`. We don't populate the set or ever get it, but do we want to make a separate enum for flags we don't track? I could also explicitly exclude it from the set creation but that seems messier

DragaDoncila: exclude it from the built sets and add a comment that in future we might want to add a `NodeCounts` or something enum that tracks flags with counts

DragaDoncila:
ok @cmalinmayor @msschwartz21 I've excluded the `min_buffer_correct` flag from the tracked sets on the `TrackingGraph`. Note I had to add a little fix in the `get_subgraph` method but that should be fine since we're removing that method anyway. I think this is now ready for a final pass.

PR #246, opened 2025-04-29
Document branching correctness = division F1 score

msschwartz21:
# Proposed Matcher or Metric Addition
- Metric

Closes #38 by documenting that BC is equivalent to the Division F1 score. I expanded the CTC metrics documentation with code samples showing how to directly calculate and retrieve each metric. I also added a new section for the CTC bio metrics describing branching correctness.

# Checklist
Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.

- [ ] I have read the developer/contributing docs.
- [ ] I have added tests for the standard test examples documented [here](https://traccuracy.readthedocs.io/en/latest/test_cases/index.html), along with end-to-end tests.
- [ ] I have checked that I maintained or improved code coverage.
- [ ] I have added benchmarking functions for my change `tests/bench.py`.
- [ ] I have added a page to the documentation with a complete description of my matcher/metric including any references.
- [ ] I have written docstrings and checked that they render correctly in the Read The Docs build (created after the PR is opened).

# Further Comments
If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #247, opened 2025-05-02
Implement cell cycle accuracy metric

### LINKED ISSUE #41: CTC Metrics: Cell cycle accuracy (CCA)
From the 2017 Ulman paper:

> CCA reflects the ability of an algorithm to discover true distribution of cell cycle lengths in a video, considering only those tracks that are both initiated and terminated by a branching event. Each such track witnesses the development of a cell from its birth until its next division, and its length, therefore, corresponds to the cell cycle length of that cell. The CCA measure is defined as: CCA = 1- maxl(|CDFr(l) - CDFgt(l)|) where CDFr and CDFgt are cumulative distribution functions of cell cycle length occurrence probabilities in the reference annotation and the computed result, respectively, adopting a common non-parametric approach to discovering dissimilarities between two sample distributions40.

40. Brown, M.R. et al. Flow-based cytometric analysis of cell cycle via simulated cell populations. PLOS Comput. Biol. 6, e1000741 (2010).

msschwartz21:
# Proposed Matcher or Metric Addition
- Metric

Closes #41 

# Checklist

- [x] I have added tests for the standard test examples documented [here](https://traccuracy.readthedocs.io/en/latest/test_cases/index.html), along with end-to-end tests.
- [x] I have added benchmarking functions for my change `tests/bench.py`.
- [x] I have added a page to the documentation with a complete description of my matcher/metric including any references.


msschwartz21:
@DragaDoncila @cmalinmayor FYI, read the docs is currently broken. It doesn't seem like I can update the github url with the new org so I've submitted a support request to RTD to see if they can do it for me.

docs/source/metrics/ctc.md
```diff
@@ -160,5 +160,19 @@ from traccuracy.metrics import DivisionMetrics
 assert ctc_matched.matching_type == "one-to-one"
 
 div_results = DivisionMetrics(max_frame_buffer=0).compute(gt_data, pred_data)
-branching_correctness = div_results["Division F1"]
+branching_correctness = div_results.results["Division F1"]
+```
+
+### Cell Cycle Accuracy (CCA)
+
+The CCA metric captures the ability of an method identify a distribution of cell
```
DragaDoncila, line 168:
```suggestion
The CCA metric captures the ability of a method to identify a distribution of cell
```

src/traccuracy/metrics/_ctc.py
```diff
@@ -253,3 +255,128 @@ def get_weighted_error_sum(
         edge_error_counts, edge_fp_weight, edge_fn_weight, edge_ws_weight
     )
     return vertex_error_sum + edge_error_sum
+
+
+class CellCycleAccuracy(Metric):
+    """The CCA metric captures the ability of an method identify a distribution of cell
```
DragaDoncila, line 261:
```suggestion
    """The CCA metric captures the ability of a method to identify a distribution of cell
```

src/traccuracy/metrics/_ctc.py
```diff
@@ -253,3 +255,128 @@ def get_weighted_error_sum(
...
+        Returns:
+            float: the cell cycle accuracy
+        """
+        # GT and pred must both contain complete cell cycles to compute this metric
+        if np.sum(gt_lengths) == 0 or np.sum(pred_lengths) == 0:
+            warnings.warn(
+                "GT and pred data do not both contain complete cell cycles. Returning CCA = 0",
+                stacklevel=2,
+            )
+            return 0
```
DragaDoncila, line 372: Should we return `np.nan` here since we can't compute the metric?

DragaDoncila (commented):
Left a couple of small comments but I'll do a more thorough review once we merge #246

msschwartz21:
This PR should be merged after #246 since it depends on some of the changes introduced to the documentation in #246.

src/traccuracy/metrics/_ctc.py
msschwartz21: Yes you're right. That's what we're doing for other metrics so we should be consistent.

src/traccuracy/metrics/_ctc.py
```diff
@@ -253,3 +255,128 @@ def get_weighted_error_sum(
...
+        Returns:
+            list[int]: a list of complete cell cycle lengths
+        """
+        lengths = []
+        # Loop over subgraphs
+        for g in track_graph.get_connected_components():
+            divs = g.get_divisions()
+            # Need at least two divisions to calculate cell cycle length
+            if len(list(divs)) >= 2:
+                lengths.extend(self._get_subgraph_lengths(g, divs))
```
DragaDoncila, line 289: I think there should be a quicker way to do this than walking the graph. I'm thinking we first make a subgraph that filters out any connected components where there are fewer than 2 nodes with outgoing degree >= 2. This should be a quick check as we just need to pass the set of nodes in the component to `out_degree` and get value counts in the list. 

This leaves us with a subgraph containing only relevant connected components. Then, we create another subgraph, removing all division-involved edges. If we track the dividing nodes from the subgraph step above, this might help us avoid a second iteration. Once we have that subgraph, we should be able to just call `len` on the edges in each connected component to get all cell cycle lengths in the graph.

DragaDoncila (commented):
@msschwartz21 sorry for the delay in review here. I still need to think about the distribution computing aspect of the code, but I think my suggestion below should speed up getting the length of each cell cycle

src/traccuracy/metrics/_ctc.py
```diff
@@ -253,3 +255,128 @@ def get_weighted_error_sum(
...
+        # GT and pred must both contain complete cell cycles to compute this metric
+        if np.sum(gt_lengths) == 0 or np.sum(pred_lengths) == 0:
+            warnings.warn(
+                "GT and pred data do not both contain complete cell cycles. Returning CCA = 0",
+                stacklevel=2,
+            )
+            return np.nan
+
+        max_track_length = np.max([np.max(gt_lengths), np.max(pred_lengths)])
+        bins = np.arange(0, max_track_length + 1)
```
DragaDoncila, line 375: Since the lengths are all integers it's faster to use `np.bincount` directly rather than working out bins and then doing `np.histogram`. There's a `min_length` arg to `bincount` that will make sure the output arrays are the same size.

src/traccuracy/metrics/_ctc.py
DragaDoncila: I realized we need to keep strictly tracklets that start and end in nodes with outgoing degree > 1. I'm going to take a look at this on Thursday. I suspect there's a memory or quadratic issue somewhere in the traversal, so I'll profile and report back.

DragaDoncila (commented):
Took another pass at this with Juan as well, I'll take a closer look later in the week and suggest some improvements.

msschwartz21:
Decisions
- Switch to `skan` for getting cell cycle lengths
- Add a test on the Hela dataset and pin to the CCA score just to detect changes
- Pull out `get_lengths` into a standalone utility function outside the class

msschwartz21 (commented):
Conclusion is we want a few more test cases

tests/metrics/test_ctc_metrics.py
```diff
@@ -161,3 +166,84 @@ def test_get_tra():
     # edge case AOGM_0 is 0
     with pytest.warns(UserWarning, match="AOGM0 is 0"):
         assert np.isnan(metrics._get_tra(errors, 0, 0))
+
+
+class TestCellCycleAccuracy:
+    cca = CellCycleAccuracy()
+
+    def get_multidiv_graph(self):
```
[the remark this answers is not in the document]
msschwartz21: Singleton nodes cause a failure if they are present regardless of whether there are any other edges in the graph.

DragaDoncila (approved):
All looks good from my end!

PR #249, opened 2025-05-12
Add function to expand skip edges

DragaDoncila:
If you are implementing a new matcher or metric, please append this `&template=new_matcher_metric.md` to your url to load the correct template.

# Proposed Change
Briefly describe the contribution. If it resolves an issue or feature request, be sure to link to that issue.

# Types of Changes
- New feature or enhancement
- Documentation update
- Tests and benchmarks

- Loaders
- Matchers
- Track Errors
- Metrics

# Checklist

# Further Comments

### DIFF
--- a/src/traccuracy/utils.py
+++ b/src/traccuracy/utils.py
@@ -1,0 +1,90 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Hashable
+
+    from traccuracy.matchers._base import Matched
+
+
+def is_equivalent_expanded_skip_edge(
+    skip_other_matched: Matched,
+    skip_src: Hashable,
+    skip_dst: Hashable,
+    matched_src: Hashable,
+    matched_dst: Hashable,
+) -> bool:
+    """True if edge skip_src -> skip_dst is equivalent to edges matched_src ->...-> matched_dst.
+
+    A skip edge skip_src -> skip_dst is equivalent to edges connecting matched_src and
+    matched_dst if:
+        - skip_src is a valid match for matched_src,
+        - skip_dst is a valid match for matched_dst,
+        - matched_src is an ancestor of matched_dst (regardless of intervening nodes),
+        - all nodes on the path matched_src -> .. -> matched_dst have incoming and outgoing
+        degrees of 1, AND
+        - all nodes on the path matched_src -> .. -> matched_dst have no valid matches in
+        skip_matched.
+
+    Parameters
+    ----------
+    skip_other_matched : Matched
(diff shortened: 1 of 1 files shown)


src/traccuracy/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        other_pred = other_predecessors[0]
+        if other_pred == matched_src:
+            # we've traversed back to matched_src with all conditions met
+            return True
+        # otherwise we have to keep traversing
+        # check that this predecessor doesn't have a match in skip graph
+        if other_pred in other_skip_map:
+            return False
+        # check that this predecessor has incoming & outgoing degree of 1
+        if other_graph.in_degree(other_pred) > 1 or other_graph.out_degree(other_pred) > 1:  # type: ignore[operator]
```
DragaDoncila, line 86: relax this check, we're ok with divisions on the path

DragaDoncila:
- Toggle will have option to only expand GT edges or only expand pred edges
- Skip-equivalent edges will have their own error annotations so users can run strict and not strict
	- add tests where it runs both ways on same graph
- Remove "expanded" from toggle names so it doesn't imply adding nodes

DragaDoncila:
@msschwartz21 yep, it would be marked as `SKIP_TP`

msschwartz21 (changes requested):
I think this is very close to ready. I left comments for a few things that I think should be addressed before merging.

I also want to make sure we have an accurate to do list for remaining skip edge tasks. Here's what I can think of, some of these may have issues and others we may need to open:
- Update basic error documentation with skip edge examples
- Implement skip edges for divisions
- Update actual metric numbers to incorporate skip annotations (basic, division, and track overlap (?))
- Updates to the gap closing edges page in the docs

Anything else that I missed?

DragaDoncila:
@msschwartz21 I think this is ready for another pass, I've resolved all review comments. One thing I noticed is that we've started using both `gap-close` terminology and `skip-edge` terminology. We should pick one and be consistent. I prefer `skip edge` because I feel like it's easier to say :laughing: but I don't have a strong preference.

> Anything else that I missed?

I think that's everything...

msschwartz21:
> One thing I noticed is that we've started using both `gap-close` terminology and `skip-edge` terminology. We should pick one and be consistent. I prefer `skip edge` because I feel like it's easier to say 😆 but I don't have a strong preference.

I prefer `skip-edge` because it just makes more intuitive sense to me.

msschwartz21 (approved):
LGTM!

I've opened a few issues to document remaining work on skip edge related things. I lumped the change to the nomenclature into the documentation issue so that we can get this one merged.

#260, #261, #262, #263

PR #259, opened 2025-06-30
Add TF measure to track_overlap

### LINKED ISSUE #40: CTC Metrics: Track Fractions (TF)
From the 2017 Ulman paper:

> Track fractions. TF targets the longest, correctly reconstructed, continuous fraction of a detected reference track. The final TF value for a particular video is computed by averaging these fractions over all detected reference tracks.

DragaDoncila:
# Proposed Change
Closes #40

This PR adds an additional result to the `TrackOverlap` metrics representing track fractions (TF): the **unweighted** track effectiveness.

This is calculated as the average fraction of maximally overlapping predicted tracklets onto the GT tracklets. Unlike track effectiveness, this metric does not take into account the tracklet length when computing fractions.

Unlike track purity/track effectiveness, it's not trivial to "invert" the track fractions number, so I only test the "forward" direction in this case. I also got a couple of floating point errors in the tests, so I changed the assertion to `np.allclose`.

# Types of Changes
- New feature or enhancement

- Metrics

# Checklist

### DIFF
--- a/src/traccuracy/metrics/_track_overlap.py
+++ b/src/traccuracy/metrics/_track_overlap.py
@@ -17,6 +17,8 @@ from collections import defaultdict
 from itertools import product
 from typing import TYPE_CHECKING, Any
 
+import numpy as np
+
 from traccuracy.matchers._base import Matched
 
 from ._base import Metric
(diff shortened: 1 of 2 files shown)


msschwartz21 (approved):
LGTM! Based on yesterday, we are going to merge this as is and then update to handle skip edges in a separate PR. 

We already have an open issue about track overlap docs so the bulk of the writing can happen later, but can you add a placeholder section to the track overlaps metrics page in the docs?

PR #290, opened 2025-07-29
Implement a geff exporter for annotated graphs and traccuracy results

### LINKED ISSUE #273: Export graphs and statistics together into a zarr with geffs and json file

msschwartz21:
Closes #273. I put the new function in utils but maybe it should live elsewhere?

~~Note: Waiting on a release from geff and the pin in pyproject.toml needs to updated to reflect that release~~

# Types of Changes
- New feature or enhancement

# Checklist

# Further Comments

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -320,5 +320,5 @@ class Test_save_results_json:
 
         # path already exists
         out_path.touch()
-        with pytest.raises(ValueError, match=f"out_path {out_path} already exists"):
+        with pytest.raises(ValueError, match=r"out_path .* already exists"):
             save_results_json([results], out_path)


msschwartz21:
@cmalinmayor @DragaDoncila I believe the tests are failing on 3.10 because of different behavior with str Enums. 

Currently the graph is annotated directly with `NodeFlag.TP_DIV` (for example). When we write the geff on python >= 3.11, the property that is written is called "NodeFlag.TP_DIV". However on python 3.10 the property that is written is "is_tp_div". 

Should we change the tests to allow either case? Or try to get the behavior to be the same?

cmalinmayor:
Garg python enums are such a headache. I feel like we should be putting the string on the graph, not the enum. However, adding `.value` everywhere we annotate also feels bad (been there, done that), so I'm open to being convinced otherwise

src/traccuracy/utils.py
```diff
@@ -136,3 +140,96 @@ def get_corrected_division_graphs_with_delta(
...
+                props_meta[str(flag)] = {
+                    "identifier": str(flag),
+                    "dtype": "bool",
+                    "description": f"Target frame buffer {target_frame_buffer}",
+                }
+            meta.node_props_metadata = props_meta
+            meta.write(geff_path)
+
+    # Write results json
+    with open(os.path.join(out_zarr, "traccuracy-results.json"), mode="w") as f:
```
DragaDoncila, line 236: I think it would be nice to extract this one-liner into its own public util that takes a path, and then call that here. That way if someone just wants to write the results they can do that. Maybe instead of util it could be a results object method? Happy with either.

DragaDoncila (commented):
Left a couple of comments @msschwartz21 , nothing too major

PR #294, opened 2025-07-30
Raise error if only one graph has error annotations

DragaDoncila:
# Proposed Change

This PR checks that error classification functions are being called with either both graphs having error annotations, or neither graph having error annotations. This ensures a user cannot end up with inconsistent annotations on graphs by accidentally re-running classification on a graph that already has errors annotated e.g. by keeping GT the same and only changing pred.

In future we may want to add a utility that "cleans" an annotated graph so that a user can reuse the graph without requiring a copy.

# Types of Changes

- Track Errors

# Checklist

### DIFF
--- a/src/traccuracy/track_errors/_ctc.py
+++ b/src/traccuracy/track_errors/_ctc.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import warnings
 from typing import TYPE_CHECKING
 
 from tqdm import tqdm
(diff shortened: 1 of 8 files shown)


src/traccuracy/track_errors/_ctc.py
```diff
@@ -69,13 +78,21 @@ def get_edge_errors(matched_data: Matched) -> None:
...
+
     if comp_graph.edge_errors and gt_graph.edge_errors:
-        logger.info("Edge errors already calculated. Skipping graph annotation")
+        warnings.warn("Edge errors already calculated. Skipping graph annotation", stacklevel=2)
         return
 
     # Node errors must already be annotated
     if not comp_graph.node_errors and not gt_graph.node_errors:
-        logger.warning("Node errors have not been annotated. Running node annotation.")
+        logger.info("Node errors have not been annotated. Running node annotation.", stacklevel=2)
```
msschwartz21, line 95: In theory we could delete these conditionals and just let the check be handled by the node and edge error functions. But it doesn't bother me that much if we leave it.

msschwartz21 (approved):

src/traccuracy/track_errors/_ctc.py
DragaDoncila: Discussed and decided to punt on this one. There are a couple of tests that do check this was logged, so we can always come back to it later.

PR #320, opened 2025-11-05
Use pylapy for solving the LAP (sparse) in PointMatcher

### LINKED ISSUE #319: Non deterministic (and incorrect?) assignment solving in PointMatcher
* traccuracy version: 0.4.2
* Python version: 3.12
* Operating System: Linux

# Description

Hey!

First of all, thanks for all this work, it's amazing. I'm integrating traccuracy in my python tracking lib ([ByoTrack](https://github.com/raphaelreme/byotrack)) and it has been quite easy to do up to this first issue.

It seems that the LAP solved in PointMatcher is non deterministic and probably slightly erroneous (similar than issue #240, though should have been fixed by a sorting of the nodes in #243). The PointMatcher depends on the labeling of the nodes, typically if one randomly shuffles and relabels the node ids (and the corresponding edges) in the nx DiGraph, the matcher does not match the same nodes, and thus the metrics are different, though metrics should be node-order agnostic.

In practice, I had this issue when building the graph from my library, as I did not iterate through the nodes in the same order as Traccuracy. And therefore I did not get the same metrics... After digging a bit, here is a minimal example to reproduce with this "dummy" node shuffling example:

```python
import random

import networkx as nx

import traccuracy
from traccuracy import run_metrics
from traccuracy.loaders import load_ctc_data
from traccuracy.matchers import CTCMatcher, IOUMatcher
from traccuracy.metrics import CTCMetrics, DivisionMetrics
from traccuracy.utils import export_graphs_to_geff


# Let's load some gt and pred tracks
ctc_folder = "..."
dataset = "DIC-C2DH-HeLa"
seq = 1
gt_data = load_ctc_data(
    f"{ctc_folder}/{dataset}/{seq:02}_GT/TRA/",
    f"{ctc_folder}/{dataset}/{seq:02}_GT/TRA/man_track.txt",
    name="Hela-01_GT",
)
pred_data = load_ctc_data(
    f"{ctc_folder}/{dataset}/{seq:02}_RES",
    f"{ctc_folder}/{dataset}/{seq:02}_RES/res_track.txt",
    name="Hela-01_RES",
)

# Compute the metrics (I'd like to use the CHOTA with a PointMatcher
results, _ = run_metrics(
    gt_data=gt_data,
    pred_data=pred_data,
    matcher=traccuracy.matchers.PointMatcher(50.0),
    metrics=[
        traccuracy.metrics.CHOTAMetric()
    ],  # CTCMetrics(), also works with CTCMetrics (because it comes from the matcher)
)

# Clear to avoid weird border effects
gt_data.clear_annotations()
pred_data.clear_annotations()

# Now, let's create a random mapping to relabel our gt graph
nodes = list(gt_data.graph.nodes)
random.shuffle(nodes)
random_mapping = {node: nodes[i] for i, node in enumerate(gt_data.graph.nodes)}
gt_random = traccuracy.TrackingGraph(
    nx.relabel_nodes(gt_data.graph, random_mapping, copy=True), location_keys=("y", "x")
)

# and compute again the metrics
rand_results, _ = run_metrics(
    gt_data=gt_random,
    pred_data=pred_data,
    matcher=traccuracy.matchers.PointMatcher(50.0),
    metrics=[
        traccuracy.metrics.CHOTAMetric()
    ],  # CTCMetrics(), also works with CTCMetrics (because it comes from the matcher)
)

# Clear to avoid weird border effects
gt_data.clear_annotations()
pred_data.clear_annotations()

# Should match, but it does not. (probably depend on the dataset and predictions, as it is random)
print(results[0]["results"], rand_results[0]["results"])

# Let's confirm that it comes from the matcher
rmap_gt_pred = dict(
    traccuracy.matchers.PointMatcher(100)._compute_mapping(
        gt_random,
        pred_data,
    )
)  # Mapping from gt_random to pred

map_gt_pred = dict(
    sorted(
        traccuracy.matchers.PointMatcher(50)._compute_mapping(
            gt_data,
            pred_data,
        )
    )
)  # Mapping from gt to pred

# Reorder the random mapping
rrmap_gt_pred = {}
for key, value in random_mapping.items():
    if value in rmap_gt_pred:
        rrmap_gt_pred[key] = rmap_gt_pred[value]

# See a part of the missmatch
for key, value in map_gt_pred.items():
    other_value = rrmap_gt_pred.get(key)
    if value != other_value:
        # print("=================================")
        print(f"Default: {key} => {value}, Shuffled: {key} => {other_value}")
        # print(
        #     gt_data.nodes[key],
        #     pred_data.nodes[value],
        #     pred_data.nodes[other_value] if other_value is not None else None,
        #     sep="\n",
        # )
```

# Fix proposition

I investigated a bit and I found that it comes from the Linear assignment problem. I think that the extension that is done in the Matcher to support non-square matrices and an alternative cost is incorrect (I did not fully check the math though). In fact, using my [pylapy](https://github.com/raphaelreme/pylapy) lib that typically solves these kind of LAPs (extension for non-square matrices and support for an alternative linking cost), the issue is solved (I checked the math there). I can do a minimal PR for this if you are interested, either switching to pylapy for the LAP solving, or just hard fixing the cost extension in the code.

With pylapy:
Simplify the code, support sparse LAP directly from the scm costs. But it add a dependency to pylapy (+ numba if sparse solving)

Hard fixing:
Minimal change (though the cost extension will stay hard to read)

Let me know if you manage to reproduce and if you are interested by a PR to fix this ;)

### Additional "bug"
I also found that with the current version, you have some link above the threshold. Adding this in the for loop of the PointMatcher will show that some selected links are indeed above threshold:
```python
if cost_matrix[row, col] > self.threshold:
    print(cost[row, col])
```

raphaelreme:
It resolves #319

PointMatcher was not solving the correct extended LAP problem. Now it does, and node labeling do not impact the metric any longer (except if two costs are strictly equals).

In addition, the LAP is solved using scipy csgraph instead of linear_sum_assignment, as the LAP is sparse, it is much faster.

I also modified IOUMatcher for its one-to-one assignments, so that it similarly relies on the sparse solving from pylapy.

- Matchers (PointMatcher, IOUMatcher)

# Checklist

### DIFF
--- a/src/traccuracy/matchers/_point.py
+++ b/src/traccuracy/matchers/_point.py
@@ -2,8 +2,8 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from scipy.spatial import KDTree
 import pylapy
+from scipy.spatial import KDTree
 
 from ._base import Matcher
 
@@ -45,10 +45,7 @@ class PointMatcher(Matcher):
         self._matching_type = "one-to-one"
 
         # Lap solver using scipy
-        self._solver = pylapy.LapSolver(
-            implementation="scipy", sparse_implementation="csgraph"
-        )
-
+        self._solver = pylapy.LapSolver(implementation="scipy", sparse_implementation="csgraph")
 
     def _compute_mapping(
         self, gt_graph: TrackingGraph, pred_graph: TrackingGraph
@@ -103,7 +100,7 @@ class PointMatcher(Matcher):
             pred_kdtree, max_distance=self.threshold, output_type="coo_matrix"
         )
 
-        # Let's keep treshold * 4 for compatibility. But one could probably do
+        # Let's keep threshold * 4 for compatibility. But one could probably do
         # hard thresholding instead (using hard=True) which is indeed what we want to do
         links = self._solver.sparse_solve(sdm, self.threshold).T
 


msschwartz21 (approved):
Thank you for the excellent contribution! 🚀

PR #346, opened 2026-05-19
Validate matching in public error annotation functions 

### LINKED ISSUE #345: _classify_divisions does not validate one-to-one mapping requirement
## Problem

`_classify_divisions` in `src/traccuracy/track_errors/_divisions.py` documents that it only works with one-to-one node mappers:

- Line 25: `"This function only works on node mappers that are one-to-one"`
- Line 38: `"ValueError: mapper must contain a one-to-one mapping of nodes"`

However, the function never actually checks the mapping type or raises the documented `ValueError`.

With many-to-one matching (multiple GT nodes → one pred node), the function may silently produce incorrect division classifications. For example, if GT node A (a division parent) and GT node B both map to pred node P, the function might classify P as a TP division based on B's lineage rather than A's.

## Suggested fix

Either:
1. Add validation at the start of `_classify_divisions` that checks `matched_data.matching_type` and raises `ValueError` for non-one-to-one mappings (as documented)
2. Or investigate whether the function actually works correctly for many-to-one cases and update the docstring accordingly

`evaluate_division_events` (the public entry point that calls `_classify_divisions`) should probably also surface this validation.

msschwartz21:
Closes #345 

# Types of Changes

- Track Errors

# Checklist

# Further Comments

### DIFF
--- a/src/traccuracy/matchers/_iou.py
+++ b/src/traccuracy/matchers/_iou.py
@@ -124,6 +124,10 @@ def _construct_time_to_seg_id_map(
         AssertionError: If two nodes in a time frame have the same segmentation_id
     """
     time_to_seg_id_map: dict[int, dict[Hashable, Hashable]] = {}
+
+    if graph.label_key is None:
+        raise ValueError("No label_key provided for input TrackingGraph")
+
     for node_id, data in graph.nodes(data=True):
         time = data[graph.frame_key]
         seg_id = data[graph.label_key]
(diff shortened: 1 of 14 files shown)


cmalinmayor (changes requested):
Can we also add a test for CompleteTracks that calling with "basic" errors on a many-to-one matching raises an error? Since this is the actual functionality that failed before.
Otherwise looks great! 🎉

cmalinmayor (approved):
👍 🌮 🍅 🍅

PR #347, opened 2026-05-27
Update references in README.md with easytrack paper

msschwartz21:


cmalinmayor (approved):
Checked the generated docs, link works, LGTM! 🎉 🦖

PR #354, opened 2026-07-07
Only return nan F1 with two input nans

DragaDoncila:
# Proposed Change

This PR changes our return values for division metrics and basic metrics so that only `Metric._get_f1(np.nan, np.nan)` returns `np.nan`.

Previously, F1 was nan if EITHER precision or recall were nan. However, this is not quite correct. Consider the following counts:

TP: 0
FP: 0
FN: 3

While `precision=np.nan`, the F1 is actually 2 * (TP=0) / 2*(TP=0 + FP=0 + FN=3) = 0/6 = 0, not nan. For F1, the only situation in which the denominator is nan, and therefore the result should be nan, is if TP==FP==FN==0, in which case both precision and recall are nan.

# Types of Changes

- Metrics

# Checklist

# Further Comments


msschwartz21 (approved):
LGTM pending pre-commit fixes

PR #355, opened 2026-07-15
Add SparseWeightedEdgeDivisionJaccard (SWEDJ): royerlab sparse-GT competition score + dense/sparse capability marker

bentaculum:
## Summary

This PR introduces a first-class notion of **sparse-ground-truth-capable metrics** in traccuracy, and adds a new metric under it.

Motivation: every existing traccuracy metric assumes **dense** ground truth (every predicted object should map to a GT object, so unmatched predictions are false positives). On **sparse** GT — where only a subset of real cells are annotated — that over-penalizes correct predictions of unannotated cells. A sparse-capable metric never penalizes unmatched predictions and can therefore also be run on dense GT; the reverse is not true.

Two parts: (1) a programmatic capability marker that classifies every metric and is surfaced in results, and (2) a new sparse metric that uses it.

## What's in here

**Sparse-GT capability marker** (`src/traccuracy/metrics/_base.py`)
- `Metric.supports_sparse_gt: bool = False` (default → all existing metrics are marked dense-only with no edits), mirrored onto the instance in `__init__` so it surfaces in `Results.metric_info`.
- Classifies the whole metric suite; the two metrics that already only score GT structure and ignore unmatched predictions — `CompleteTracks` and `CompleteTracksByLength` — are opted in alongside the new one.
- Docs gain a "Dense vs. sparse ground truth" section explaining the distinction, listing the sparse-capable metrics, and grounding the concept in the literature — the sparse-annotation lineage-completeness eval of [Malin-Mayor et al., *Nat Biotechnol* 2023](https://doi.org/10.1038/s41587-022-01427-7) (basis for Complete Tracks by Length) and the CZ Biohub competition (SWEDJ). Complete Tracks' sparse tolerance is stated as a computed property (not a CTC claim), with a caveat that within-track errors (e.g. a division with an unannotated daughter) are still counted, and a note that the marker is per-metric-object (so the mixed Track Overlap Metrics is marked dense-only despite its sparse-tolerant TE/TF).

**New sparse metric — `traccuracy.metrics.SparseWeightedEdgeDivisionJaccard` (SWEDJ)** (`src/traccuracy/metrics/_swedj.py`), the royerlab competition's edge + division Jaccard
- One-to-one centroid-distance node matching, reusing the existing **`PointMatcher`** (`threshold`, `scale_factor` for anisotropy) — no new matcher.
- **Edge Jaccard**: a predicted edge is TP when both endpoints match GT nodes joined by a GT edge; FP only when there is GT evidence against it (source matched to a GT node with an out-edge, or target to one with an in-edge); everything else is ignored (this is the sparse rule). `FN = |GT edges| − TP`.
- **Adjusted edge Jaccard**: optional excess-node penalty via `n_gt_nodes` (coarse estimate of the true node count incl. unannotated cells).
- **Division Jaccard**: ±1-frame-tolerant; per-division *local* re-matching + connected-component/one-node-stage/daughter-lineage coverage + bipartite fork pairing; FP clamp from the global match.
- **Combined `score` = adj_edge_jaccard + 0.1·division_jaccard**. Per-pair output; sum the `*_tp/*_fp/*_fn` across pairs to micro-average a dataset score (matches the competition aggregation).

**Docs** — new `metrics/swedj.md` (with the edge/division-Jaccard illustrations reused from the competition under BSD-3, credited + upstream `LICENSE` retained under `docs/source/metrics/figures/`); the metrics overview now has a "Dense vs. sparse ground truth" section, a `Sparse GT` column, and the blanket dense-assumption warning now points at the per-metric marker.

## Parity / testing

`tests/metrics/test_swedj.py` pins parity with the reference implementation via the hand-computed **edge-Jaccard unit cases** ported from the upstream test suite (spurious edges, cross-track, reparenting, self-loops, false merges, skip connections, etc.), plus a small set of **minimal division fixtures in the repo's own style** (`tests/examples/graphs.py`) covering division TP / FN / FP-with-clamp / late-within-±1-tolerance / two-independent-divisions. Marker, adjusted-Jaccard, score, node-recall, empty-pred and input-validation tests round it out. All new tests pass; ruff + mypy clean; the existing in-memory metric/track-error suite still passes.

> Note: during development the port was validated against all 14 of the competition's frozen sandbox counts; the committed suite trades that verbatim import for concise, house-style examples.

## Open questions for review

1. **Name** — `SparseWeightedEdgeDivisionJaccard` (SWEDJ), module `_swedj.py` (covers both the edge and division Jaccard).
2. **Requires `PointMatcher`** — the metric raises `TypeError` for other matchers, since division scoring re-matches by centroid distance. Reasonable, or should it degrade to the global match for other one-to-one matchers?
3. **Marker polarity** — `supports_sparse_gt` default `False` (chosen so existing metrics need no edits) vs. a negative `dense_only`.
4. **`score` without `n_gt_nodes`** — now falls back to the raw `edge_jaccard` term (so the metric returns a usable score by default); `adj_edge_jaccard`/`total_node_ratio` stay `NaN` to signal the excess-node penalty was skipped. Prefer this, or a hard `NaN` to force supplying the estimate?


bentaculum:
Superseded by https://github.com/live-image-tracking-tools/traccuracy/pull/356, which hosts the branch on this repo directly (same-repo PR) so CI has the write access the coverage job needs. Same branch and commits.

PR #356, opened 2026-07-16
Add supports_sparse_gt capability marker for sparse ground-truth metrics

bentaculum:
## Summary

Introduces a first-class notion of **sparse-ground-truth-capable metrics** in traccuracy.

Every existing metric assumes **dense** ground truth (every real cell is annotated), so a predicted node/edge with no ground-truth match is a false positive. On **sparse** ground truth — only a subset of cells annotated — that over-penalizes correct predictions of unannotated cells. This PR adds a programmatic capability marker so metrics can be classified accordingly.

## What's in here

- **`Metric.supports_sparse_gt: bool = False`** (`src/traccuracy/metrics/_base.py`) — default `False` (dense-only), mirrored onto the instance in `__init__` so it surfaces in `Results.metric_info`. A sparse-capable metric does not treat a prediction as an error *solely* for lacking a ground-truth counterpart, and can always be run on dense GT too; the reverse is not true.
- **Classifies the existing suite**: the two metrics that already only score ground-truth structure and ignore unmatched predictions — `CompleteTracks` and `CompleteTracksByLength` — are marked sparse-capable; all others inherit dense-only with no edits.
- **Docs**: a "Dense vs. sparse ground truth" section and a `Sparse GT` column in the metrics overview.
- **Tests** (`tests/metrics/test_base.py`): default polarity, opt-in via subclass, surfacing in `Results`, and the classification of the existing metrics.

A new sparse metric that uses this marker — **SparseWeightedEdgeDivisionJaccard (SWEDJ)** — is stacked on top of this PR in a follow-up.


msschwartz21 (commented):
I haven't had a chance to go through this fully, but I think we may need a slightly more nuanced classification of metrics than just True/False at the metric level. For example, anything that uses false negatives would still be valid on sparse data which means that for most of our Metric classes, some of the computed values are valid on sparse and others are not.

msschwartz21:
@bentaculum I've started another approach to this PR that will allow key level classification of sparse support. I should have the PR open and ready for review on Monday.

bentaculum:
Closing in favor of #363.

PR #357, opened 2026-07-17
Add SparseWeightedEdgeDivisionJaccard (SWEDJ) sparse tracking metric

bentaculum:
## Summary

Adds **`SparseWeightedEdgeDivisionJaccard` (SWEDJ)** — a traccuracy implementation of the [royerlab cell-tracking-competition](https://github.com/royerlab/kaggle-cell-tracking-competition) score (edge Jaccard + division Jaccard) for **sparsely annotated** ground truth.

> **Stacked on #356** (the `supports_sparse_gt` capability marker). Base branch: `feat/sparse-tracking-metric`. Review/merge #356 first; this diff is SWEDJ-only.

## What's in here

**`traccuracy.metrics.SparseWeightedEdgeDivisionJaccard`** (`src/traccuracy/metrics/_swedj.py`), marked `supports_sparse_gt = True`
- One-to-one centroid-distance node matching, reusing the existing **`PointMatcher`** (`threshold`, `scale_factor`) — no new matcher.
- **Edge Jaccard**: a predicted edge is TP when both endpoints match GT nodes joined by a GT edge; FP only when there is GT evidence against it; everything else ignored (the sparse rule). `FN = |GT edges| − TP`.
- **Adjusted edge Jaccard**: optional excess-node penalty via `n_gt_nodes`.
- **Division Jaccard**: ±1-frame-tolerant; per-division *local* re-matching + connected-component/one-node-stage/daughter-lineage coverage + bipartite fork pairing; FP clamp.
- **Combined `score` = adj_edge_jaccard + 0.1·division_jaccard** (falls back to raw `edge_jaccard` when `n_gt_nodes` is absent). Per-pair output; sum counts across pairs to micro-average a dataset score.

**Docs** — new `metrics/swedj.md` (edge/division-Jaccard figures reused from the competition under BSD-3, credited + upstream `LICENSE` retained under `docs/source/metrics/figures/`) + a row in the metrics overview.

## Parity / testing

`tests/metrics/test_swedj.py` pins parity with the reference implementation via hand-computed edge-Jaccard unit cases ported from the upstream test suite, plus minimal division fixtures (TP / FN / FP-with-clamp / late-within-±1 / two-divisions) in the repo's own style. ruff + mypy clean.

## Open questions

1. **Name** — `SparseWeightedEdgeDivisionJaccard` (SWEDJ), module `_swedj.py`.
2. **Requires `PointMatcher`** — raises `TypeError` for other matchers (division scoring re-matches by centroid distance). Reasonable, or degrade to the global match?
3. **`score` without `n_gt_nodes`** — falls back to raw `edge_jaccard`; prefer that or a hard `NaN`?


bentaculum:
Needs rebasing once #363 lands.

msschwartz21:
@bentaculum finally ready for refactoring this PR now that #363 is merged in.

docs/source/metrics/swedj.md
```diff
@@ -0,0 +1,123 @@
...
+)
+```
+
+## Edge Jaccard
+
+Predicted edges are first hardened against implausible topology: only edges spanning a
+single forward timepoint ($t_{target} = t_{source} + 1$) are scored — backward,
+same-frame and gap-closing edges are dropped — and each source keeps at most two
+outgoing edges (a division has at most two children), so surplus children are dropped
+too. A surviving predicted edge $(u, v)$ is a **true positive** when both endpoints are
```
msschwartz21, line 44: This is more of a high level question, but is there any penalty for edges with implausible topology or is this similar to the ctc where a bunch of edges are dropped without any penalty prior to actual evaluation?

src/traccuracy/metrics/_swedj.py
```diff
@@ -0,0 +1,568 @@
...
+    # counts (agnostic). Nothing here is dense-only.
+    sparse_safe_keys = frozenset(
+        {
+            "edge_tp",
+            "edge_fp",
+            "edge_fn",
+            "edge_jaccard",
+            "division_tp",
+            "division_fp",
+            "division_fn",
```
msschwartz21, line 117: I'm wondering if we should change the nomenclature for the edge_fp/fn and division_fp/fn terms since they are computed in a distinct way from the normal edge and division metrics. Maybe just prefixing the terms with swedj_

msschwartz21 (commented):
I started reviewing this before I realized it's still in draft mode so I'll just share the two initial comments that I had from looking through the docs. I didn't actually get into the real metric code

PR #358, opened 2026-07-21
Add load_napari_data loader

bentaculum:
## Summary

Adds `load_napari_data`, a loader that converts an in-memory napari Tracks layer into a `TrackingGraph` so napari tracking results can be matched and evaluated with traccuracy — mirroring `load_point_data` but for the napari format.

The loader takes the plain `.data` / `.graph` / `.properties` arrays and dicts of a napari Tracks layer (not a layer object), so **traccuracy gains no dependency on napari**.

- `data` is `[track_id, t, (z), y, x]` per detection; `graph` is `{child_track_id: [parent_track_id, ...]}`.
- Within-track edges connect consecutive detections of the same `track_id` ordered by time.
- Cross-track edges connect each parent track's last detection to each child track's first detection.
- Optional `segmentation` + `seg_id_key` wire up `segmentation_id` node attributes for CTC-style matching.

## Notes

- Merges (a child track with >1 parent) are rejected — CTC-style evaluation doesn't support them.
- Times must be integer-valued; a bare-int or list parent in `graph` is accepted, as is a numpy scalar/0-d array.

## Testing

- `tests/loaders/test_napari.py` — 13 tests covering 2D/3D, divisions (list, scalar, and numpy-scalar parents), merge rejection, duplicate `(track_id, t)`, segmentation wiring, seg-id length mismatch, time gaps, and non-integer-time rejection.

### DIFF
--- a/src/traccuracy/loaders/_napari.py
+++ b/src/traccuracy/loaders/_napari.py
@@ -42,7 +42,8 @@ def load_napari_data(
             with columns ``[track_id, t, (z), y, x]``. ``D`` is 2 or 3.
         graph (Mapping[int, Sequence[int]] | None, optional): The napari Tracks
             ``graph``, mapping each child track id to its parent track id(s).
-            Defaults to None (no divisions).
+            The parent may be a bare int or a list. Defaults to None (no
+            divisions).
         properties (Mapping[str, Sequence] | None, optional): Per-detection
             properties (same length/order as ``data`` rows), e.g. the napari
             Tracks layer ``properties``. Used to read segmentation label ids.
@@ -138,6 +139,9 @@ def load_napari_data(
     # child track's first detection. napari graph is {child: [parents]}.
     if graph:
         for child_track, parent_tracks in graph.items():
+            # napari allows a bare int or a list of parent track ids.
+            if np.isscalar(parent_tracks):
+                parent_tracks = [parent_tracks]
             if len(parent_tracks) > 1:
                 raise ValueError(
                     f"track {child_track} has multiple parents "
--- a/src/traccuracy/loaders/_napari.py
+++ b/src/traccuracy/loaders/_napari.py
@@ -91,7 +91,7 @@ def _labels_by_matching(
     data: np.ndarray,
     segmentation: np.ndarray,
     ndim: int,
-    progbar_class=tqdm,
+    progbar_class: type[tqdm] = tqdm,
 ) -> np.ndarray:
     """Assign each detection a segmentation label by per-frame optimal matching.
 
@@ -183,7 +183,7 @@ def load_napari_data(
     segmentation: np.ndarray | None = None,
     seg_id_key: str | None = None,
     name: str | None = None,
-    progbar_class=tqdm,
+    progbar_class: type[tqdm] = tqdm,
 ) -> TrackingGraph:
     """Load a napari Tracks layer into a TrackingGraph.
 
--- a/src/traccuracy/loaders/_napari.py
+++ b/src/traccuracy/loaders/_napari.py
@@ -6,8 +6,8 @@ from typing import TYPE_CHECKING
 
 import networkx as nx
 import numpy as np
-from scipy.ndimage import center_of_mass
 from scipy.optimize import linear_sum_assignment
+from tqdm import tqdm
 
 from traccuracy._tracking_graph import TrackingGraph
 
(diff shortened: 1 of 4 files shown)


src/traccuracy/loaders/_napari.py
```diff
@@ -0,0 +1,174 @@
...
+    :class:`~traccuracy.TrackingGraph` so it can be matched/evaluated, mirroring
+    :func:`load_point_data` but for the in-memory napari format.
+
+    Edges are built two ways: consecutive detections of the same ``track_id``
+    (ordered by time) are connected, and each parent track's last detection is
+    connected to each child track's first detection from ``graph``.
+
+    This function takes plain arrays/dicts (the ``.data`` / ``.graph`` /
+    ``.properties`` of a napari Tracks layer) rather than a layer object, so
+    traccuracy does not depend on napari.
```
msschwartz21, line 156: The design here of grabbing the plain data instead of the layer is an elegant way to avoid napari as a dependency. Can we include a code snippet example of how to use it for someone who isn't as familiar with napari?

bentaculum: Good call — added an `Example:` section to the docstring in 07a3f23. It shows the common case (passing `tracks_layer.data/.graph/.properties` straight from a napari Tracks layer), a raw-array version with a division for anyone building the arrays by hand, and the segmentation variant for CTC-style matching.

msschwartz21 (approved):

bentaculum:
@msschwartz21 — summary of what's changed since you added the labels, all in `load_napari_data`:

**Implicit segmentation matching (the main addition).** You can now pass just a `segmentation` and skip the `seg_id_key`/`properties` bookkeeping entirely — each detection's label is read directly from the pixel under its `(t, (z), y, x)` position (`segmentation[t, (z), y, x]`), assuming the point sits inside its own mask. The explicit `seg_id_key` path stays as the opt-in for when positions don't land cleanly on their masks. The docstring now leads with a two-mode (implicit vs. explicit) callout so the choice is visible up front.

- I benchmarked the implicit read against the explicit one: it's ~15–110x slower *relative to* a bare column read, but that's a ratio on a trivially cheap op — in absolute terms it's <1% of total loader time even at 50k detections (the graph construction and `TrackingGraph.__init__` dominate). No meaningful cost.

**Merges are now allowed.** Dropped the old "a child track can't have multiple parents" rejection — a child with several parents now builds a merge node (in-degree ≥ 2), which `TrackingGraph` already supports via `get_merges()`. The input graph can represent merges.

**Docs reframed.** User-facing references to matching were CTC-specific; they're now "segmentation-based" since both the CTC and IoU matchers consume `segmentation_id`.

**Edge-case hardening (from an internal review pass).** Empty-data + segmentation no longer throws a cryptic numpy error; non-integer track ids are rejected (mirroring the existing time check) so float ids can't silently collapse tracks; and the explicit label column is validated integer-valued instead of silently truncating (`6.9 → 6` matched the wrong mask).

Net: `src/.../loaders/_napari.py` +136 / tests +83, 22 tests total.

One thing I pulled back out and will send as its own PR: general self-loop / cycle validation. A track listed as its own parent (`{1: [1]}`) creates a cycle, and I found that traccuracy doesn't enforce acyclicity *anywhere* — not in `TrackingGraph`, not in the other loaders. That's a general `TrackingGraph` concern rather than something to special-case in one loader, so it's out of scope here and coming separately.

msschwartz21:
@bentaculum Thanks for the recap of changes! I'll take a more careful pass tomorrow. It also occurred to me that it may make sense to add this function to our benchmarks.

bentaculum:
@msschwartz21 Good idea — added `load_napari_data` to `tests/bench.py` in cbfb5ce, mirroring `test_load_points`:

- `test_load_napari` — bare data load (300 detections),
- `test_load_napari_implicit_seg` — the implicit segmentation-matching path (labels read by position).

For reference, at 300 detections the implicit-segmentation load runs at essentially the same speed as the bare load (mean ~1.04 ms vs ~1.04 ms), confirming the by-position label read adds no meaningful overhead on top of graph construction.

bentaculum:
✍️ @msschwartz21 I propose to make the implicit matching between masks and tracks pretty permissve, and therefore extended it to a linear assignment problem between center of mass from segmentations and track edges anchor points, to cover edge cases like track edge anchors being just outside of cells e.g. due to highly non-convex mask or slight manual misplacement.

bentaculum:
For some context, this loader is needed for the [divisualisation napari plugin](github.com/bentaculum/divisualisation) for calculating errors directly via GUI. I am hosting it directly there for now, and can switch over to the hardened version in here once it is part of a traccuracy release.

bentaculum:
@msschwartz21 I made some changes since you approved, most importantly the implicit matching with LAP, so tagged you for review once more.

src/traccuracy/loaders/_napari.py
```diff
@@ -0,0 +1,424 @@
...
+        raise ValueError(
+            f"segmentation has {segmentation.ndim} dims but data implies a "
+            f"{ndim}D image plus time ({ndim + 1} dims); cannot match labels "
+            "from positions. Pass precomputed labels via seg_id_key instead."
+        )
+
+    if len(data) == 0:
+        return np.empty(0, dtype=int)
+
+    t_idx = data[:, 1].astype(np.intp)
```
msschwartz21: The 1 index here caught my eye because it didn't match the (t,z,y,x) order described in line 19. Not sure if this is a documentation issue or an actual bug?

src/traccuracy/loaders/_napari.py
```diff
@@ -0,0 +1,424 @@
...
+        total=len(unique_times),
+        desc="Matching detections to masks",
+        leave=False,
+    ):
+        rows = np.nonzero(times == t)[0]
+        frame = segmentation[int(t)]
+        # Mask labels and centers of mass, in the same (z,)y,x coordinate order
+        # as data[:, 2:].
+        labels, coms = _mask_centroids(frame)  # labels (M,), coms (M, ndim)
+        if len(labels) == 0:
```
msschwartz21: Should this be `if len(labels) == 0 and len(rows) != 0`?

src/traccuracy/loaders/_napari.py
```diff
@@ -0,0 +1,424 @@
...
+
+    if seg_id_key is not None and segmentation is None:
+        raise ValueError(
+            "seg_id_key was given without segmentation; pass a segmentation to "
+            "enable label matching, or drop seg_id_key."
+        )
+
+    # Resolve each detection's segmentation label id (or None if no matching).
+    # Two modes when a segmentation is given: explicit (read from a properties
+    # column) or, by default, implicit (read the pixel under each position).
```
msschwartz21:
```suggestion
    # column) or, by default, implicit (match detection positions to the centroid of each label).
```

src/traccuracy/loaders/_napari.py
```diff
@@ -0,0 +1,424 @@
...
+from scipy.optimize import linear_sum_assignment
+from tqdm import tqdm
+
+from traccuracy._tracking_graph import TrackingGraph
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping, Sequence
+
+
+def _labels_from_positions(data: np.ndarray, segmentation: np.ndarray, ndim: int) -> np.ndarray:
```
msschwartz21: Should this be deleted since it's been replaced by the matching version?

msschwartz21 (approved):
A few minor comments/clarifications, but otherwise this looks good! Good call with the matching. Sorry it took me a bit to get back to you!

Let me know what you think about the comments I left, but I'll approve now so you're not blocked once they are addressed.

src/traccuracy/loaders/_napari.py
bentaculum: Yep, good catch — removed it in d208f08. It was the old pixel-lookup matcher, fully superseded by the centroid LAP one.

src/traccuracy/loaders/_napari.py
bentaculum: Not a bug — col 1 is t (data is [track_id, t, (z), y, x], so t is index 1, not 0). Moot now anyway, that whole function's gone in d208f08.

src/traccuracy/loaders/_napari.py
bentaculum: Thx, applied — and fixed up the other spots in the docstring that still said "pixel under position" too (d208f08).

src/traccuracy/loaders/_napari.py
bentaculum: Right, done in 43616cd. The loop only visits times that appear in the data so len(rows) is always >0 today, but the guard reads clearer this way and won't misfire if that ever changes.

msschwartz21:
@bentaculum I think this is ready to merge if you are happy!

bentaculum:
Yes I'm a happy merger now

PR #359, opened 2026-07-21
Validate that TrackingGraph edges go forward in time

bentaculum:
## Summary

`TrackingGraph` is documented as "a directed graph representing a tracking solution where edges go forward in time," but nothing enforced it. Self-loops, same-frame edges, and cycles all passed through construction silently — even though downstream code assumes strictly forward edges (`is_skip_edge` computes `frame[src] + 1 != frame[dst]`, the complete-tracks metric computes `target_frame - source_frame` expecting a positive span, and the matchers rely on the DAG/forward structure).

This came up while reviewing the napari loader (#358): a self-parent entry `{1: [1]}` would build a 2-cycle that no layer caught. Rather than special-case it in one loader, this enforces the invariant centrally where it belongs.

## Change

Adds `TrackingGraph._validate_edge`, called per edge in `_set_attrs` under the existing `validate` flag. Every edge `(src, dst)` must satisfy `frame[dst] > frame[src]`. This:

- rejects **self-loops** (`frame[dst] == frame[src]`),
- rejects **backward** and **same-frame** edges,
- **guarantees acyclicity** — since the frame strictly increases along every edge, no cycle can exist,
- still **allows skip edges** (spanning more than one frame — they go forward),
- and is **bypassed by `validate=False`**, consistent with the existing node validation. (`clear_annotations()` already calls `_set_attrs(validate=False)`, so it's unaffected.)

Uses `assert` to match the style of the sibling `_validate_node` rather than mixing in a `ValueError`.

## Fixture fixes

Four pre-existing tests built degenerate graphs with same-frame edges that only worked because nothing validated them. I verified each is accidental (not testing same-frame behavior) and that no production code path legitimately produces such edges — the CTC public loader, for instance, already rejects a daughter that starts before its parent ends:

- **`nx_comp1_pos_list`** — its edge list was a copy-paste error vs. the sibling `nx_comp1` (a self-loop `(3,3)` and same-frame `(3,5)`); restored to the intended edges.
- **`test_ctc_single_nodes`** — a daughter tracklet started in its parent's final frame (invalid CTC); moved it one frame later.
- **`test_assign_edge_errors`** — all nodes were at `t=0`; edge sources now sit in frame 0, targets in frame 1.
- **`test_simple_2d` (geff)** — the synthetic geff sample graph is non-temporal (paired nodes share a frame); reassigns a unique increasing time per node so the loaded graph is forward-in-time.

## Testing

- 6 new tests: self-loop / backward / same-frame all raise; forward + skip edges pass; merge (in-degree 2) passes; `validate=False` bypasses.
- Full suite: 661 passed, no regressions. ruff / ruff-format / mypy / typos clean.

Independent design review by Codex confirmed the acyclicity argument, the skip-edge compatibility, and the `assert`-over-`ValueError` consistency call.

### DIFF
--- a/src/traccuracy/_tracking_graph.py
+++ b/src/traccuracy/_tracking_graph.py
@@ -286,6 +286,8 @@ class TrackingGraph:
 
         # store edge id in edges_by_flag
         for edge, attrs in self.graph.edges.items():
+            if validate:
+                self._validate_edge(edge)
             for edge_flag in EdgeFlag:
                 if attrs.get(edge_flag.value):
                     self.edges_by_flag[edge_flag.value].add(edge)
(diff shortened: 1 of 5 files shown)


msschwartz21 (approved):
LGTM!

PR #363, opened 2026-08-10
Classify sparse-ground-truth validity per result key, not per metric

### LINKED ISSUE #333: Support for sparse ground truth
# Description
Add a flag to the ground truth graph that indicates whether or not it is sparse. Then, only compute metrics that support sparse-ness, or have a sparse implementation. For example:
- Don't report FPs in the error counts
- Weight FPs zero in the summary metrics (AOGM, TRA, CHOTA, etc.)

If we initially error when running non-supported metrics, we can add support incrementally.

# Topics

- New feature or enhancement

- Track Errors
- Metrics

# Priority
- [x] This is an essential feature
- [ ] Nice to have
- [ ] Future idea

- [x] Yes! :tada:

msschwartz21:
## Summary

This PR will replace #356 if everyone is happy with it. That PR classifies sparse-ground-truth validity with one `supports_sparse_gt: bool` per `Metric` class. Working through the existing metrics turned up several where a single `_compute()` call mixes sparse-safe and dense-only outputs -- a class-level flag can't express that:

- `DivisionMetrics` returns a sparse-safe `Division Recall` alongside a dense-only `Division Precision` from the same call.
- `TrackOverlapMetrics` returns sparse-safe `target_effectiveness`/`track_fractions` alongside dense-only `track_purity`.
- `BasicMetrics` mixes sparse-safe TP/FN/recall counts with dense-only FP/precision/F1 counts, per node and edge.

This PR reworks the capability marker to classify **per result key** instead:

- `Metric.sparse_safe_keys` / `Metric.agnostic_keys`: class-level sets a metric uses to opt in specific output keys. `agnostic_keys` covers raw counts (e.g. `Total GT Nodes`) that are accurate regardless of annotation density but aren't themselves a quality judgment -- neither "safe" nor "unsafe". Anything not listed in either set defaults to dense-only (the previous default posture, preserved).
- `Metric._classify_sparse_safe(key)`: classifies a single key as `"sparse_safe"`, `"agnostic"`, or `"dense_only"`.
- `Metric.compute(..., sparse_only=True)`: filters a metric's results down to its sparse-safe/agnostic keys, recursing into nested per-key structures (e.g. `DivisionMetrics`' per-frame-buffer buckets) rather than misclassifying the bucket key itself. If filtering leaves nothing (e.g. `CTCMetrics`, whose errors are already collapsed into single dense-only scores), it warns instead of silently returning `{}`.
- `sparse_safe_keys`/`agnostic_keys` surface (as sorted tuples) in `Results.metric_info`, same JSON-serializable pattern as the original PR.

Docs updated with a per-metric breakdown table of sparse-safe/agnostic keys, replacing the class-level Yes/No column.

Closes #333 by adding a is_sparse_gt flag to the TrackingGraph. Metrics check that flag in compute and set sparse_only accordingly if it's not already set.

## Test plan

- [x] `pixi run test tests/metrics/` -- 381 passed
- [x] Full suite: 697 passed (1 pre-existing, unrelated `test_geff.py` failure)
- [x] `ruff check`, `mypy` clean
- [x] Per-metric end-to-end tests via `compute(sparse_only=True)`: flat-dict filtering (`test_basic.py`), nested frame-buffer filtering (`test_divisions.py`), dense-only-subset filtering (`test_track_overlap_metrics.py`), the "no safe keys at all" warning case (`test_ctc_metrics.py`), and the "nothing filtered" no-op case (`test_complete_tracks.py`, `test_complete_tracks_by_length.py`)

### DIFF
--- a/docs/source/metrics/metrics.md
+++ b/docs/source/metrics/metrics.md
@@ -36,6 +36,10 @@ Each `Metric` classifies every key in the dictionary it returns, via `Metric._cl
 - **agnostic**: a raw count or other value that is not itself a quality judgment (e.g. `Total GT Nodes`). Accurate regardless of annotation density, so no claim is made either way.
 - **dense-only** (the default for any key not listed above): counts or derives from unmatched predictions (false positives, precision, F1, ...) and will over-penalize correct predictions of unannotated ground truth.
 
+:::{warning}
+Although we classify certain metrics as `sparse-safe`, this does not mean that they, in isolation, provide meaningful insight into the true overall accuracy of the solution. One can trivially imagine, for example, a solution with an overwhelming number of annotations that achieves perfect node recall. These `sparse-safe` metrics also provide no insight on the quality or diversity of the ground truth annotations present. We advise extreme caution in interpreting or reporting these metrics.
+:::
+
 Sparseness is a property of the annotations rather than a per-run choice, so it is declared on
 the ground truth graph itself and is read-only after construction. Pass `is_sparse_gt=True` to
 {class}`~traccuracy.TrackingGraph` or to any loader (`load_ctc_data`, `load_geff_data`,
--- a/docs/source/metrics/metrics.md
+++ b/docs/source/metrics/metrics.md
@@ -14,8 +14,8 @@ they can accept, and a brief description of behavior and any hyperparameters.
 
 Many metrics support relaxing skip edges for the ground truth and/or the prediction. Relaxing a skip edge means allowing one edge that spans multiple frames to match multiple edges in the other graph, thus potentially reducing the number of errors.
 
- Metric category | Matching Type(s) | Description |
-------------------|------------------|-------------
+| Metric category | Matching Type(s) | Description |
+|------------------|------------------|-------------|
 | [Basic Metrics](basic-metrics): TP, FP, and FN nodes and edges | `one-to-one`  | Counts the number of **true positive** (matched) nodes and edges, **false positive** (unmatched in the prediction) nodes and edges, and **false negative** (unmatched in the ground truth) nodes and edges. |
 | [Division metrics](division-metrics): TP, FP, and FN divisions and F1 score/Branching Correctness (BC) | `one-to-one` | Counts the number of **true positive** (matched) divisions, **false positive** (unmatched in the prediction) divisions, and **false negative** (unmatched in the ground truth) divisions. Then computes the division **F1-Score**, also called **Branching Correctness** by the CTC-Bio metrics. Has a `max_frame_buffer` parameter that allows counting divisions as correct within `max_frame_buffer` frames as long as the parent and children match within the buffer| 
 | [CTC Metrics](ctc-metrics): DET, LNK, TRA | `one-to-one`, `many-to-one` | A set of three metrics between 0 and 1, with higher scores indicating better performance. DET measures node errors, LNK measures linking errors, and TRA combines detection and linking errors. |
@@ -36,9 +36,25 @@ Each `Metric` classifies every key in the dictionary it returns, via `Metric._cl
 - **agnostic**: a raw count or other value that is not itself a quality judgment (e.g. `Total GT Nodes`). Accurate regardless of annotation density, so no claim is made either way.
 - **dense-only** (the default for any key not listed above): counts or derives from unmatched predictions (false positives, precision, F1, ...) and will over-penalize correct predictions of unannotated ground truth.
 
+Sparseness is a property of the annotations rather than a per-run choice, so it is declared on
+the ground truth graph itself and is read-only after construction. Pass `is_sparse_gt=True` to
+{class}`~traccuracy.TrackingGraph` or to any loader (`load_ctc_data`, `load_geff_data`,
+`load_napari_data`, `load_point_data`), and every metric computed against that graph
+restricts itself to sparse-safe and agnostic keys:
+
+```python
+gt = load_ctc_data("path/to/gt", is_sparse_gt=True)
+```
+
 :::{note}
-`Metric.compute` has a `sparse_safe` option which can be set to True in order to return only sparse safe results.
+`Metric.compute` also takes a `sparse_only` option, for filtering against a graph that was not
+marked sparse at construction. A graph marked `is_sparse_gt=True` always filters, regardless of
+this option.
 :::
+
+A metric that declares no sparse-safe or agnostic keys at all (CCA, CHOTA) cannot report
+anything under sparse ground truth. Those are skipped with a warning rather than computed and
+discarded.
 
 Sparse-safe and agnostic keys by metric (everything else that metric returns is dense-only):
 
@@ -46,7 +62,7 @@ Sparse-safe and agnostic keys by metric (everything else that metric returns is 
 |---|---|---|
 | [Basic Metrics](basic-metrics) | `True/False Negative {Nodes,Edges}`, `{Node,Edge} Recall`, `Skip GT/Pred True Positive Edges`, `Skip False Negative Edges` | `Total GT/Pred {Nodes,Edges}` |
 | [Division metrics](division-metrics) | `Division Recall`, `True/False Negative Divisions`, `Wrong Children Divisions`, `True Positive Skip Divisions` | `Total GT Divisions`, `Total Predicted Divisions` |
-| [CTC Metrics](ctc-metrics) | none | none |
+| [CTC Metrics](ctc-metrics) | `fn_nodes`, `fn_edges` | none |
 | [Cell Cycle Accuracy (CCA)](cca) | none | none |
 | [Complete Tracklets and Lineages](complete-tracks) | `correct_lineages`, `correct_tracklets`, `complete_lineages`, `complete_tracklets` | `total_lineages`, `total_tracklets` |
 | [Track Overlap Metrics](track-overlap-metrics) | `target_effectiveness`, `track_fractions` | none |
--- a/docs/source/metrics/metrics.md
+++ b/docs/source/metrics/metrics.md
@@ -32,7 +32,7 @@ Most metrics assume **dense** ground truth (every real cell is annotated), so a 
 
 Each `Metric` classifies every key in the dictionary it returns, via `Metric._classify_sparse_safe(key)` (also surfaced as `sparse_safe_keys`/`agnostic_keys` on {class}`~traccuracy.metrics.Results`'s `metric` info). Many metrics mix categories: for example {class}`~traccuracy.metrics.DivisionMetrics` reports both a sparse-safe recall and a dense-only precision from the same call. The three categories are:
 
-- **sparse-safe**: judges only structure the annotation can judge (matches, ground-truth-only counts, false negatives). Valid on both sparse and dense ground truth.
+- **sparse-safe**: a quality/error assessment that does not assume predictions with no match are false positives so it remains valid when ground truth is sparse.
 - **agnostic**: a raw count or other value that is not itself a quality judgment (e.g. `Total GT Nodes`). Accurate regardless of annotation density, so no claim is made either way.
 - **dense-only** (the default for any key not listed above): counts or derives from unmatched predictions (false positives, precision, F1, ...) and will over-penalize correct predictions of unannotated ground truth.
 
--- a/docs/source/metrics/metrics.md
+++ b/docs/source/metrics/metrics.md
@@ -14,9 +14,33 @@ they can accept, and a brief description of behavior and any hyperparameters.
 
 Many metrics support relaxing skip edges for the ground truth and/or the prediction. Relaxing a skip edge means allowing one edge that spans multiple frames to match multiple edges in the other graph, thus potentially reducing the number of errors.
 
-:::{warning}
-Unless otherwise noted, metrics are written assuming that you have dense ground truth annotations. The results on sparse annotations may be unpredictable and should be interpreted cautiously.
+(sparse-annotations)=
+## Dense vs. sparse ground truth
+
+Most metrics assume **dense** ground truth (every real cell is annotated), so a predicted node, edge, or division with no ground truth match is a false positive. On **sparse** ground truth, where only a subset of cells are annotated, that would over-penalize correct predictions of unannotated cells.
+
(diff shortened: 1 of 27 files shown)


docs/source/metrics/metrics.md
```diff
@@ -14,17 +14,57 @@ they can accept, and a brief description of behavior and any hyperparameters.
...
+the ground truth graph itself and is read-only after construction. Pass `is_sparse_gt=True` to
+{class}`~traccuracy.TrackingGraph` or to any loader (`load_ctc_data`, `load_geff_data`,
+`load_napari_data`, `load_point_data`), and every metric computed against that graph
+restricts itself to sparse-safe and agnostic keys:
+
+```python
+gt = load_ctc_data("path/to/gt", is_sparse_gt=True)
+```
+
+:::{note}
```
cmalinmayor, line 49: Confused about this one

docs/source/metrics/metrics.md
```diff
@@ -14,17 +14,57 @@ they can accept, and a brief description of behavior and any hyperparameters.
...
+| Metric | Sparse-safe keys | Agnostic keys |
+|---|---|---|
+| [Basic Metrics](basic-metrics) | `True/False Negative {Nodes,Edges}`, `{Node,Edge} Recall`, `Skip GT/Pred True Positive Edges`, `Skip False Negative Edges` | `Total GT/Pred {Nodes,Edges}` |
+| [Division metrics](division-metrics) | `Division Recall`, `True/False Negative Divisions`, `Wrong Children Divisions`, `True Positive Skip Divisions` | `Total GT Divisions`, `Total Predicted Divisions` |
+| [CTC Metrics](ctc-metrics) | `fn_nodes`, `fn_edges` | none |
+| [Cell Cycle Accuracy (CCA)](cca) | none | none |
+| [Complete Tracklets and Lineages](complete-tracks) | `correct_lineages`, `correct_tracklets`, `complete_lineages`, `complete_tracklets` | `total_lineages`, `total_tracklets` |
+| [Track Overlap Metrics](track-overlap-metrics) | `target_effectiveness`, `track_fractions` | none |
+| [Complete Tracks by Length](complete-tracks-by-length-metric) | `correct`, `accuracy` | `total` |
+| [Cell-specific Higher Order Tracking Accuracy (CHOTA)](chota-metric) | none | none |
```
cmalinmayor, line 68: How hard to auto compute this table.... 🤔

src/traccuracy/loaders/_ctc.py
```diff
@@ -253,6 +254,10 @@ def load_ctc_data(
         border_margin (float, optional): If set, nodes whose centroid is within this
             distance (in pixels) of the spatial border will be excluded from the graph.
             Defaults to None (no filtering).
+        is_sparse_gt (bool, optional): Set to True if this is ground truth with sparse
```
cmalinmayor, line 257: Should this be "is_sparse"? technically predictions could also be sparse and we could do an extension for that at some point in the future... but I am not sold that's worth it.

src/traccuracy/metrics/_base.py
```diff
@@ -22,8 +22,32 @@ class Metric(ABC):
...
+    predicted node/edge/division with no ground truth match is treated as a false
+    positive. On **sparse** ground truth, where only a subset of cells are
+    annotated, that over-penalizes correct predictions of unannotated cells.
+
+    Subclasses can classify each key in the dict returned by ``_compute`` (matched
+    by name, regardless of how deeply it is nested, e.g. under a per-frame-buffer
+    bucket) via ``sparse_safe_keys`` and ``agnostic_keys``:
+
+    - ``sparse_safe_keys``: the key is a quality/error assessment that only judges
+      structure the annotation can judge (matches, ground-truth-only counts, false
```
cmalinmayor: This phrasing I don't love.... how about "does not assume predictions with no match are false positives"

src/traccuracy/metrics/_base.py
```diff
@@ -53,6 +77,83 @@ def _validate_matcher(self, matched: Matched) -> bool:
...
+            str: One of "sparse_safe", "agnostic", or "dense_only". See the class
+                docstring for what each means.
+        """
+        if key in type(self).sparse_safe_keys:
+            return "sparse_safe"
+        if key in type(self).agnostic_keys:
+            return "agnostic"
+        return "dense_only"
+
+    #: Substrings (matched case-insensitively) identifying sparse-safe keys that are still
```
cmalinmayor, line 98: Add this to the docs? And the PR template for new metrics, so that it gets updated if new metrics are added?

docs/source/metrics/metrics.md
cmalinmayor: After reading implementation it seems like this is the same mechanism that is_sparse_gt=True uses. Perhaps the second sentence should be: "A graph marked 'is_sparse_gt=True' always sets this to True."? Or... idk but this phrasing is confusing 🤷‍♀️

docs/source/metrics/metrics.md
```diff
@@ -14,17 +14,57 @@ they can accept, and a brief description of behavior and any hyperparameters.
...
+
+(sparse-annotations)=
+## Dense vs. sparse ground truth
+
+Most metrics assume **dense** ground truth (every real cell is annotated), so a predicted node, edge, or division with no ground truth match is a false positive. On **sparse** ground truth, where only a subset of cells are annotated, that would over-penalize correct predictions of unannotated cells.
+
+Each `Metric` classifies every key in the dictionary it returns, via `Metric._classify_sparse_safe(key)` (also surfaced as `sparse_safe_keys`/`agnostic_keys` on {class}`~traccuracy.metrics.Results`'s `metric` info). Many metrics mix categories: for example {class}`~traccuracy.metrics.DivisionMetrics` reports both a sparse-safe recall and a dense-only precision from the same call. The three categories are:
+
+- **sparse-safe**: judges only structure the annotation can judge (matches, ground-truth-only counts, false negatives). Valid on both sparse and dense ground truth.
```
cmalinmayor: Same comment as in the docstring - This phrasing is weird to me.

docs/source/metrics/metrics.md
cmalinmayor: Could even do without the note TBH, the normal workflow is just the graph annotation

msschwartz21: Does this rephrasing help?

Metrics computed on a graph that is marked as `is_sparse_gt=True` at creation will always use the `sparse_only=True` option. However sparse mode can always be enabled by manually setting the `sparse_only=True` flag when calling `Metric.compute`.

tests/metrics/test_sparse_safe_keys.py
```diff
@@ -0,0 +1,193 @@
+"""Pin each metric's declared sparse-safe/agnostic keys to the keys it actually returns.
+
+``sparse_safe_keys``/``agnostic_keys`` repeat result-key strings that ``_compute`` builds
+independently -- ``BasicMetrics`` composes them with f-strings and never writes them as
+literals, and ``DivisionMetrics`` writes them out twice. Nothing ties the two together, and
+the failure is silent: an unrecognized key is classified ``dense_only`` and quietly dropped
+from sparse results, while a declared key that no longer exists is a dead no-op. These tests
+turn both into a test failure.
```
cmalinmayor, line 8: ❓

tests/metrics/test_sparse_safe_keys.py
```diff
@@ -0,0 +1,193 @@
...
+        produced |= _leaf_keys(metric_factory()._compute(matched_factory(), **compute_kwargs))
+
+    assert declared <= produced, (
+        f"declared but never returned: {sorted(declared - produced)}. A renamed or "
+        "misspelled key is silently treated as dense-only and dropped from sparse results."
+    )
+
+
+@pytest.mark.parametrize("metric_class", ALL_METRIC_CLASSES, ids=lambda c: c.__name__)
+def test_sparse_safe_and_agnostic_are_disjoint(metric_class):
```
cmalinmayor, line 141: Ooh I think I get it, these are tests that assert the keys the subclasses implement are valid. I do feel like these should be somehow in the __init__ as like a post-init step? Or otherwise error at runtime and not just in the test suite? But I assume you thought about this 🤔

tests/metrics/test_sparse_safe_keys.py
```diff
@@ -0,0 +1,193 @@
...
+
+def test_metrics_without_declared_keys_are_skipped_not_computed():
+    # CCA and CHOTA declare nothing, so under sparse ground truth they cannot report
+    # anything. Guards the list in the docs table against silently going stale.
+    for metric_class in (CellCycleAccuracy, CHOTAMetric):
+        assert not metric_class.sparse_safe_keys
+        assert not metric_class.agnostic_keys
+
+
+def test_every_metric_subclass_is_covered():
```
cmalinmayor, line 155: Also in developer docs or PR template for new metrics 🙏

tests/metrics/test_sparse_safe_keys.py
```diff
@@ -0,0 +1,193 @@
...
+    g.add_edge(1, 2)
+    gt = TrackingGraph(g.copy(), location_keys=("y", "x"), is_sparse_gt=True)
+    pred = TrackingGraph(g.copy(), location_keys=("y", "x"))
+    matched = Matched(gt, pred, [(1, 1), (2, 2)], {"matching type": "one-to-one"})
+
+    with pytest.warns(UserWarning, match="GT graph is marked is_sparse_gt=True"):
+        results = BasicMetrics().compute(matched).results
+
+    declared = BasicMetrics.sparse_safe_keys | BasicMetrics.agnostic_keys
+    assert set(results.keys()) <= declared
```
cmalinmayor, line 191: <=? isn't it just equal?

tests/metrics/test_sparse_safe_keys.py
```diff
@@ -0,0 +1,193 @@
...
+        sub
+        for c in subclasses
+        for sub in c.__subclasses__()
+        if sub.__module__.startswith("traccuracy")
+    }
+    uncovered = sorted(c.__name__ for c in subclasses - set(ALL_METRIC_CLASSES))
+    assert not uncovered, f"uncovered Metric subclasses: {uncovered}"
+
+
+def test_leaf_keys_recurses():
```
cmalinmayor, line 169: This test is testing a function that is also defined in the tests....

tests/metrics/test_sparse_safe_keys.py
```diff
@@ -0,0 +1,193 @@
...
+        lambda: CompleteTracks(error_type="ctc"),
+        [(larger_example_1, {})],
+        id="CompleteTracks-ctc",
+    ),
+    pytest.param(
+        lambda: CompleteTracksByLength(max_length=2),
+        [(larger_example_1, {})],
+        id="CompleteTracksByLength",
+    ),
+    pytest.param(
```
cmalinmayor, line 109: I'm very confused by this test TBH. Should the tests for each metric assert this independently rather than having a parameterized one here that doesn't even use the ALL_METRIC_CLASSES helper?

cmalinmayor (changes requested):
Mostly documentation and test confusion from me, the actual implementation seems solid. Apologies if some of these were covered in the meeting, my brain was not recording at that time.

The main concrete request is to add whatever needs to be done when a new metric is added to the developer docs/PR template for a new metric. Preferrably this is as simple as possible, but if the test parameterization is actually adding complexity rather than removing it, I think it's fine to say "Add a test that the sparse metrics are correctly computed to your metric test suite".

docs/source/metrics/metrics.md
msschwartz21: Not easy enough that I want to spend time on it now. We have lots of tables in the docs that have to be manually maintained and kept up to date

src/traccuracy/loaders/_ctc.py
msschwartz21: Personally I think explicit is better and we can generalize to predictions in the future

tests/metrics/test_sparse_safe_keys.py
msschwartz21: @bentaculum can you respond to the comments/questions on this file?

docs/source/metrics/metrics.md
```diff
@@ -14,17 +14,55 @@ they can accept, and a brief description of behavior and any hyperparameters.
...
+
+Each `Metric` classifies every key in the dictionary it returns, via `Metric._classify_sparse_safe(key)` (also surfaced as `sparse_safe_keys`/`agnostic_keys` on {class}`~traccuracy.metrics.Results`'s `metric` info). Many metrics mix categories: for example {class}`~traccuracy.metrics.DivisionMetrics` reports both a sparse-safe recall and a dense-only precision from the same call. The three categories are:
+
+- **sparse-safe**: a quality/error assessment that does not assume predictions with no match are false positives so it remains valid when ground truth is sparse.
+- **agnostic**: a raw count or other value that is not itself a quality judgment (e.g. `Total GT Nodes`). Accurate regardless of annotation density, so no claim is made either way.
+- **dense-only** (the default for any key not listed above): counts or derives from unmatched predictions (false positives, precision, F1, ...) and will over-penalize correct predictions of unannotated ground truth.
+
```
DragaDoncila, line 38: Here I would like to add a warning admonition that highlights that `sparse-safe` does **not** mean that the metric can't be misleading e.g. Ben's example of detecting a bunch of stuff and eventually hitting perfect "recall" e.g.:

"Although we classify certain metrics as `sparse-safe`, this does **not** mean that they, in isolation, provide meaningful insight into the true overall accuracy of the solution. One can trivially imagine, for example, a solution with an overwhelming number of annotations that achieves perfect node recall. The metrics also provide no insight on the quality or diversity of the ground truth annotations present. We advise extreme caution in interpreting or reporting these metrics."

DragaDoncila (commented):
Had a look a this and generally like the shape of it. Also played with it locally and works a treat.
Only other thing I noticed, aside from my comment, is that if both sol and gt are marked as "is_sparse_gt", we don't warn or anything. I think that's probably ok? It doesn't materially change anything I don't think.

docs/source/metrics/metrics.md
bentaculum: Agreed this seems a good place for this disclaimer

DragaDoncila: @msschwartz21 all looks good to me, but I'd like to add a stronger disclaimer admonition here, as above.

msschwartz21: Sorry I missed this!

DragaDoncila (approved):
Thank you @msschwartz21 you're the best. Let's get this in 🧸

PR #364, opened 2026-08-13
Immutable is_sparse_gt, loader support, and tighten the sparse-key mechanism

### LINKED ISSUE #333: Support for sparse ground truth

bentaculum:
Stacked on #363 (base is `feat/sparse-gt-key-classification`, so the diff here is only my additions). Happy to squash any of this into #363 instead if that's easier, @msschwartz21 — or drop the bits you disagree with.

The per-key model in #363 is the right primitive: a class-level bool genuinely can't express `DivisionMetrics` returning a sparse-safe recall and a dense-only precision from one call. This just tightens the edges around it.

## `is_sparse_gt` is read-only, and loaders can set it

Sparseness describes the annotations, not a per-run preference — so `sparse_only=False` shouldn't be able to override it (it can't, and that's correct), and by the same logic it shouldn't be mutable after construction either. Flipping it later silently changes the meaning of results already computed from that graph. It's now backed by a private attribute behind a read-only property.

The flip side: if the graph flag is *the* route to sparse metrics, it has to be reachable where graphs are actually built. None of the loaders exposed it, so ground truth loaded the canonical way had to be patched after the fact. `is_sparse_gt` is now a parameter on `load_ctc_data`, `load_geff_data`, `load_napari_data` and `load_point_data`:

```python
gt = load_ctc_data("path/to/gt", is_sparse_gt=True)
```

I did *not* add `sparse_only` to `run_metrics`. `relax_skips_*` is a genuine analysis choice while sparseness is a data property, so the asymmetry seems right — but say the word and I'll add it.

## Applicability is decided before `_compute`, not after

Whether anything survives filtering is knowable from the class alone: if a metric declares neither set, every leaf is dense-only by construction. Moving that check ahead of `_compute` fixes three things at once.

**The "no sparse-safe keys" warning was misfiring.** It was inferred from an empty *output*, so any metric whose `_compute` legitimately returns `{}` got blamed. Reproducible on `CompleteTracksByLength`, which declares three keys:

```
WARN: GT graph has no frame range (empty graph). Returning empty results.
WARN: CompleteTracksByLength has no sparse-safe or agnostic keys, so sparse_only=True
      filtered every result out. This metric is not meaningful on sparse ground truth.
declared sparse_safe_keys: ['accuracy', 'correct'] agnostic: ['total']
```

**CCA and CHOTA no longer compute and discard.** They previously ran a full `_compute` under a sparse GT graph and then threw away 100% of it.

**`_is_empty_result` is gone** — the static check subsumes it.

Observable behaviour is unchanged: same warning, same empty results, just decided earlier. I deliberately kept it a *warning* rather than raising. #333 proposed erroring on unsupported metrics and there's a real argument for it, but that would turn a mixed `run_metrics([BasicMetrics(), CHOTAMetric()])` call on sparse ground truth into a hard failure, and that's your call rather than mine. Easy switch if you want it.

## CTC/AOGM false-negative counts are sparse-safe

You asked for a second pair of eyes on the classifications, so: `CTCMetrics(...).compute(...).results` actually returns

```
['AOGM', 'DET', 'LNK', 'TRA', 'fn_edges', 'fn_nodes', 'fp_edges', 'fp_nodes', 'ns_nodes', 'ws_edges']
```

`fn_nodes` and `fn_edges` come off `data.gt_graph` (`NodeFlag.CTC_FALSE_NEG` / `EdgeFlag.CTC_FALSE_NEG`), so they only judge objects the annotation covers — and `BasicMetrics` already declares its equivalent `False Negative Nodes`/`Edges` sparse-safe. The "already collapsed into single dense-only scores" rationale holds for TRA/DET/LNK/AOGM but not for those two raw counts, so `sparse_only=True` was dropping usable numbers and warning the metric was meaningless.

`ns_nodes` and `ws_edges` I left dense-only — they're flagged on the prediction and I wasn't confident either way.

## The declared key sets are now pinned to reality

This is the one I'd most like your opinion on, since it's about the mechanism rather than a specific classification.

`sparse_safe_keys`/`agnostic_keys` repeat result-key strings that `_compute` builds independently. `BasicMetrics` never writes `"Node Recall"` as a literal at all — it composes it via `f"{feature_type} Recall"` after `feature_type.capitalize()`. `DivisionMetrics` writes its ten keys out twice. Nothing ties any of that to the declared sets, and the failure is silent both ways: an unrecognized key is classified `dense_only` and quietly vanishes from sparse results, and a declared key that no longer exists is a dead no-op.

`tests/metrics/test_sparse_safe_keys.py` asserts `declared <= returned` per metric, unioned over the relaxed/non-relaxed cases needed to emit the conditional skip keys, plus that the two sets are disjoint and that every `Metric` subclass is covered. Verified it actually fails — renaming `"Node Recall"` to `"Nod Recall"` gives:

```
AssertionError: declared but never returned: ['Nod Recall']. A renamed or misspelled key
is silently treated as dense-only and dropped from sparse results.
```

I considered making the keys an enum like `NodeFlag`/`EdgeFlag`, or a `TypedDict` per result so mypy checks the *producing* side. Both are better guarantees, but the enum needs `BasicMetrics._compute_stats`'s f-string composition unrolled into explicit `NODE_RECALL`/`EDGE_RECALL` members, which is churn in the most-used metric and unrelated to sparse support. Feels like its own PR — and one that pays for itself on `_divisions.py`'s duplicated literals regardless.

## Docs

The note named a `sparse_safe` option; the parameter is `sparse_only`. Also documented `is_sparse_gt` as the primary route, added the loader mention, recorded CTC's fn counts in the table, and restored the leading pipe that got clipped from the metrics table header when the old warning block came out.

## One thing not addressed

#363 says "Closes #333", but #333 asks for two things and this stack delivers one. "Don't report FPs in the error counts" is done. "Weight FPs zero in the summary metrics (AOGM, TRA, CHOTA, etc.)" isn't, and filtering structurally can't deliver it — the sparse-correct TRA isn't "no TRA", it's TRA computed with `vertex_fp_weight=0`, which is a change to the computation rather than to which keys come back. `AOGMMetrics` already takes those weights as constructor args so the path is unusually short, though whether a reweighted TRA should still be called TRA is a real question. Suggest `Refs #333` and leaving it open.

## Test plan

- `pytest` — **730 passed**, 0 failures (`pixi -e dev`, py3.11 linux). Includes 33 new tests. Note I did not reproduce the pre-existing `test_geff.py` failure mentioned in #363.
- `ruff check src tests` and `ruff format --check` — clean.
- `mypy src/traccuracy` — 3 errors, all pre-existing in `_cca.py` / `_compute_overlap.py`, none in changed files.
- Mutation-checked the new contract test as shown above.

## CI

Two follow-up commits here, plus one thing that can't be fixed from this PR.

- **`fix(docs)`** — mine. I had documented `is_sparse_gt` both as a `@property` and in the class docstring's `Attributes:` list, so autoapi emitted a duplicate object description and Read the Docs (which runs sphinx with `-W`) failed. The property keeps its own docstring; `nodes`/`edges` are handled the same way and are deliberately absent from that list. Duplicate-object warnings 2 → 0 locally, and RTD is green again.
- **`test: ignore the NumPy 2.5 shape deprecation raised via skan`** — **not related to this PR's subject.** `skan` calls `skimage.util.map_array`, which still assigns to `ndarray.shape`; NumPy >= 2.5 deprecates it and `filterwarnings = ["error"]` turns it into 2 failures in `test_cca.py` and 2 in `bench.py`. Only visible on py3.12/3.13 because the resolver picks NumPy 2.5 there. It breaks `main` equally — main's last green run predates the NumPy 2.5 release — so it's dependency drift, not a code change. With it, the entire 3.12/3.13 matrix went green.
- **Benchmark stays red until #365 merges.** The job does `git checkout ${{ github.event.pull_request.base.sha }}` and benchmarks the *base* commit, which doesn't carry the ignore, and the cache key is the base SHA so it re-fails every run. It's structurally unfixable from this branch. I opened #365 against `main` with just that one commit; once it lands (and this branch's base picks it up) Benchmark should pass. Happy to drop the duplicate commit from here and rebase at that point.

### DIFF
--- a/docs/source/metrics/metrics.md
+++ b/docs/source/metrics/metrics.md
@@ -14,8 +14,8 @@ they can accept, and a brief description of behavior and any hyperparameters.
 
 Many metrics support relaxing skip edges for the ground truth and/or the prediction. Relaxing a skip edge means allowing one edge that spans multiple frames to match multiple edges in the other graph, thus potentially reducing the number of errors.
 
- Metric category | Matching Type(s) | Description |
-------------------|------------------|-------------
+| Metric category | Matching Type(s) | Description |
+|------------------|------------------|-------------|
 | [Basic Metrics](basic-metrics): TP, FP, and FN nodes and edges | `one-to-one`  | Counts the number of **true positive** (matched) nodes and edges, **false positive** (unmatched in the prediction) nodes and edges, and **false negative** (unmatched in the ground truth) nodes and edges. |
 | [Division metrics](division-metrics): TP, FP, and FN divisions and F1 score/Branching Correctness (BC) | `one-to-one` | Counts the number of **true positive** (matched) divisions, **false positive** (unmatched in the prediction) divisions, and **false negative** (unmatched in the ground truth) divisions. Then computes the division **F1-Score**, also called **Branching Correctness** by the CTC-Bio metrics. Has a `max_frame_buffer` parameter that allows counting divisions as correct within `max_frame_buffer` frames as long as the parent and children match within the buffer| 
 | [CTC Metrics](ctc-metrics): DET, LNK, TRA | `one-to-one`, `many-to-one` | A set of three metrics between 0 and 1, with higher scores indicating better performance. DET measures node errors, LNK measures linking errors, and TRA combines detection and linking errors. |
@@ -36,9 +36,25 @@ Each `Metric` classifies every key in the dictionary it returns, via `Metric._cl
 - **agnostic**: a raw count or other value that is not itself a quality judgment (e.g. `Total GT Nodes`). Accurate regardless of annotation density, so no claim is made either way.
 - **dense-only** (the default for any key not listed above): counts or derives from unmatched predictions (false positives, precision, F1, ...) and will over-penalize correct predictions of unannotated ground truth.
 
+Sparseness is a property of the annotations rather than a per-run choice, so it is declared on
+the ground truth graph itself and is read-only after construction. Pass `is_sparse_gt=True` to
+{class}`~traccuracy.TrackingGraph` or to any loader (`load_ctc_data`, `load_geff_data`,
+`load_napari_data`, `load_point_data`), and every metric computed against that graph
+restricts itself to sparse-safe and agnostic keys:
+
+```python
+gt = load_ctc_data("path/to/gt", is_sparse_gt=True)
+```
+
 :::{note}
-`Metric.compute` has a `sparse_safe` option which can be set to True in order to return only sparse safe results.
+`Metric.compute` also takes a `sparse_only` option, for filtering against a graph that was not
+marked sparse at construction. A graph marked `is_sparse_gt=True` always filters, regardless of
+this option.
 :::
+
+A metric that declares no sparse-safe or agnostic keys at all (CCA, CHOTA) cannot report
+anything under sparse ground truth. Those are skipped with a warning rather than computed and
+discarded.
 
 Sparse-safe and agnostic keys by metric (everything else that metric returns is dense-only):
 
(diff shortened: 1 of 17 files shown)


msschwartz21 (approved):
This looks great! Really good catches 😅 I'll let you merge in case you had any other lingering things you were adding.

PR #367, opened 2026-08-25
Validate declared sparse-safe/agnostic keys at class definition

bentaculum:
Stacked on #363. Follows up on [cmalinmayor's comment](https://github.com/live-image-tracking-tools/traccuracy/pull/363#discussion_r-sparse-safe-keys) on `tests/metrics/test_sparse_safe_keys.py`:

> I do feel like these should be somehow in the `__init__` as like a post-init step? Or otherwise error at runtime and not just in the test suite?

Agreed, for the half of it that can move. It splits cleanly:

- **Whether a declaration is self-consistent** depends only on the class body, so it belongs at class-definition time. Moved to `Metric.__init_subclass__`.
- **Whether declared keys match what `_compute` returns** fundamentally needs data — `BasicMetrics` composes its keys at runtime with `f"{feature_type} Recall"` after `.capitalize()`, so nothing about them is knowable from the class body. Stays a test.

## What changed

`Metric.__init_subclass__` now raises when a subclass is *defined*, so a bad declaration is an import-time error rather than something that only surfaces if a test remembered to look:

- `ValueError` if a key is in both `sparse_safe_keys` and `agnostic_keys` (`_classify_sparse_safe` silently resolves such a key to sparse-safe).
- `TypeError` if either is not a `frozenset`. This one is load-bearing rather than fussy: a mutable `set` can be edited into an overlap *after* the check has run, so immutability is what makes the disjointness check mean anything. Rebinding the attribute entirely is still unchecked — this is a definition-time check, not an invariant.

Both use the MRO-resolved values, so `CTCMetrics` is validated against what it inherits from `AOGMMetrics`.

## Why `ALL_METRIC_CLASSES` goes away

Once the hook enforces disjointness, no subclass at any depth can escape it, and `ALL_METRIC_CLASSES` had no consumer left except the test asserting the list was complete — purely circular. So it and `test_sparse_safe_and_agnostic_are_disjoint` are deleted.

The completeness guard moves to the registry that *does* still need one. Nothing forced a new metric into `DECLARED_KEY_CASES`, so its keys would never have been pinned against `_compute`. The replacement:

- walks subclasses **recursively** instead of exactly two levels (the old walk was structurally blind below `Metric → AOGMMetrics → CTCMetrics`, which is as deep as the tree currently goes — so this is future-proofing, not a live bug);
- keeps the census the old test provided via an explicit `KEYLESS_METRICS`, so a metric that declares *nothing at all* is still flagged instead of quietly defaulting to dense-only.

Net effect for a contributor adding a metric: one list to touch instead of two, and a loud error instead of a silently-skipped test.

## Also trimmed the file

Two of the five tests left in `tests/metrics/test_sparse_safe_keys.py` weren't earning their place, and cmalinmayor flagged one of them directly:

- `test_leaf_keys_recurses` only exercised this file's own flattening helper — the test suite talking to itself. Deleted; every `DECLARED_KEY_CASES` case already runs that helper for real.
- `test_metrics_without_declared_keys_are_skipped_not_computed` never checked skipping despite its name, and hardcoded the same `(CellCycleAccuracy, CHOTAMetric)` pair that `KEYLESS_METRICS` lists a few lines below. It wasn't redundant, though — it was really the *reverse* direction of the census: a metric listed as keyless that later starts declaring keys would keep being waved through while its keys went unpinned. That check moved into `test_every_metric_is_pinned_or_explicitly_keyless` and now reads `KEYLESS_METRICS` instead of a second copy of the pair.

`KEYLESS_METRICS` also moves up beside `DECLARED_KEY_CASES` so both registries can be read together. Five tests down to three.

## Test plan

- [x] Full suite: 723 passed. The one failure, `tests/loaders/test_geff.py::test_load_rel_obj`, is pre-existing on #363's head and unrelated (geff `RelatedObject.node_prop` API drift).
- [x] `pre-commit run --files <changed>`: typos, ruff, ruff-format, mypy all pass.
- [x] Mutation-tested the three new tests in `test_base.py` — no-op'ing the hook fails all three; dropping only the type check fails only `test_mutable_key_set_is_rejected`.
- [x] Confirmed the new census guard fires on both a keyed and a keyless injected metric, and that the old two-level walk missed an injected level-3 subclass the recursive one catches.
- [x] Checked for test-ordering leakage: at the moment the guard runs in a full session, the only non-traccuracy `Metric` subclasses alive are the test-local ones, all rejected by the module filter. Passes in both file orderings.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/new_matcher_metric.md
+++ b/.github/PULL_REQUEST_TEMPLATE/new_matcher_metric.md
@@ -14,7 +14,7 @@ Put an x in the boxes that apply. You can also fill these out after creating the
 - [ ] I have added a page to the documentation with a complete description of my matcher/metric including any references.
 - [ ] I have written docstrings and checked that they render correctly in the Read The Docs build (created after the PR is opened).
 - [ ] For a new metric, I have specified `sparse_safe_keys`, `agnostic_keys` and `_sparse_inflatable_substrings` in the Metric definition.
-- [ ] For a new metric, I have added it to `ALL_METRIC_CLASSES` in tests/metrics/test_sparse_safe_keys.py
+- [ ] For a new metric, I have added a `DECLARED_KEY_CASES` entry in tests/metrics/test_sparse_safe_keys.py, or listed it in `KEYLESS_METRICS` there if it has no sparse-safe or agnostic keys
 
 # Further Comments
 If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...
--- a/src/traccuracy/metrics/_base.py
+++ b/src/traccuracy/metrics/_base.py
@@ -41,12 +41,48 @@ class Metric(ABC):
     - Anything not listed in either set is treated as dense-only: it counts or
       derives from unmatched predictions (false positives, precision, F1, ...) and
       will over-penalize correct predictions of unannotated ground truth.
+
+    The two sets must be disjoint frozensets; ``__init_subclass__`` checks that when
+    the subclass is defined.
     """
 
     #: See the "sparse-safe" bullet in the class docstring.
     sparse_safe_keys: frozenset[str] = frozenset()
     #: See the "agnostic" bullet in the class docstring.
     agnostic_keys: frozenset[str] = frozenset()
+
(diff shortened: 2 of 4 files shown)


msschwartz21 (approved):
LGTM! Thanks for those improvements. Sorry the review took so long.

